Issue #9, opened 2022-06-22
dbt fails to run on duckdb==0.4.0


jecolvin:
Any dbt command (I personally tried `dbt seed` and `dbt run`) fails with the following error:

```
Runtime Error
  TransactionContext Error: cannot commit - no transaction is active
```

Downgrading to duckdb v0.3.2 resolves the issue.

jwills:
Well that's lame, I will take a look shortly as I was hoping to do the upgrade to 0.4.0 in the next version!

dbeatty10:
I got the same error message as @jecolvin with duckdb 0.4.0.

Downgrading to duckdb 0.3.4 worked again.

There's some interesting discussion about this error message here:
https://github.com/duckdb/duckdb/issues/3395

I'm guessing some default implementation(s) within dbt-core are adding `commit`s that duckdb 0.4.0 can't handle.

_Maybe_ [here](https://github.com/dbt-labs/dbt-core/blob/3e3ecb1c3f49efcfee9dbd09374bcf23e4e5b565/core/dbt/adapters/sql/connections.py#L133-L134)?

A handful of other possibilities come up via a grep or [search in GitHub](https://github.com/dbt-labs/dbt-core/search?q=commit) (but there's a bit of noise to sort through).

jwills:
So FWIW the new testing framework makes it really simple to see where things are going wrong here:

```
/usr/local/lib/python3.9/site-packages/dbt/tests/fixtures/project.py:389: in create_test_schema
    self.adapter.create_schema(relation)
/usr/local/lib/python3.9/site-packages/dbt/adapters/sql/impl.py:163: in create_schema
    self.commit_if_has_connection()
/usr/local/lib/python3.9/site-packages/dbt/adapters/base/impl.py:191: in commit_if_has_connection
    self.connections.commit_if_has_connection()
/usr/local/lib/python3.9/site-packages/dbt/adapters/base/connections.py:274: in commit_if_has_connection
    self.commit()
/usr/local/lib/python3.9/site-packages/dbt/adapters/sql/connections.py:158: in commit
    self.add_commit_query()
/usr/local/lib/python3.9/site-packages/dbt/adapters/sql/connections.py:134: in add_commit_query
    return self.add_query("COMMIT", auto_begin=False)
/usr/local/lib/python3.9/site-packages/dbt/adapters/sql/connections.py:70: in add_query
    cursor.execute(sql, bindings)
```

[...]

PR #10, opened 2022-06-24
Exclude duckdb 0.4.0 from compatible versions

dbeatty10:
Related to #9

Currently, an install of dbt-duckdb==1.1.1 in a fresh environment will yield versions like this:
- dbt-duckdb-1.1.1
- duckdb-0.4.0

This PR simply constrains the upper bound of duckdb so it will install duckdb-0.3.x instead.


jwills:
Thank you @dbeatty10! Finally have a little time to dig in and see what's going on in here.

Issue #23, opened 2022-10-14
Support for `s3_endpoint`


edgarrmondragon:
S3-compatible services require

```sql
SET s3_endpoint='<domain>.<tld>:<port>';
```

https://duckdb.org/docs/extensions/httpfs#configuration

PR #24, opened 2022-10-14
Add `s3_endpoint` parameter for MinIO and other S3-compatible services

edgarrmondragon:
Closes https://github.com/jwills/dbt-duckdb/issues/23

### DIFF
--- a/dbt/adapters/duckdb/connections.py
+++ b/dbt/adapters/duckdb/connections.py
@@ -33,6 +33,7 @@ class DuckDBCredentials(Credentials):
     s3_access_key_id: Optional[str] = None
     s3_secret_access_key: Optional[str] = None
     s3_session_token: Optional[str] = None
+    s3_endpoint: Optional[str] = None
 
     @property
     def type(self):
@@ -82,6 +83,8 @@ class DuckDBConnectionWrapper:
                 cursor.execute(
                     f"SET s3_session_token = '{credentials.s3_session_token}'"
                 )
+            if credentials.s3_endpoint is not None:
+                cursor.execute(f"SET s3_endpoint = '{credentials.s3_endpoint}'")
 
         self._cursor = DuckDBCursorWrapper(cursor)
 
--- a/README.md
+++ b/README.md
@@ -24,7 +24,8 @@ so, ya know, don't do that.
 
 As of version 1.2.0, dbt-duckdb also allows you to configure your S3 settings in your credentials, including `s3_region` and
 either `s3_session_token` or `s3_access_key_id` and `s3_secret_access_key`, so that you can use dbt-duckdb to read and transform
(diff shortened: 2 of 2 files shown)


dbt/adapters/duckdb/connections.py
```diff
@@ -82,6 +83,8 @@ def __init__(self, conn, credentials):
                 cursor.execute(
                     f"SET s3_session_token = '{credentials.s3_session_token}'"
                 )
+            if credentials.s3_endpoint is not None:
```
jwills, line 86: Do you still need credentials set (tokens or keys or w/e) if you're using a non-AWS s3 endpoint? I essentially require it right now, along with the `s3_region`-- is that right for e.g. MinIO?

edgarrmondragon: Yeah, MinIO needs username and password, and it uses the respective `s3_access_key_id` and `s3_secret_access_key` settings. The region is optional I think, but MinIO can also be configured with one.

This example adapted from the duckdb repo works fine:

```bash
docker run \
   --rm \
   -p 9000:9000 \
   -p 9090:9090 \
   --name minio \
   --hostname duckdb-minio.com \
   -v $PWD/minio/data:/data \
   -e "MINIO_REGION_NAME=eu-west-1" \
   -e "MINIO_ROOT_USER=ROOTNAME" \
   -e "MINIO_ROOT_PASSWORD=CHANGEME123" \
   -e "MINIO_DOMAIN=duckdb-minio.com" \
   quay.io/minio/minio server /data --console-address ":9090"
```

```sql
INSTALL httpfs;
LOAD httpfs;
SET s3_region='eu-west-1';
SET s3_endpoint='duckdb-minio.com:9000';
SET s3_access_key_id='ROOTNAME';
SET s3_secret_access_key='CHANGEME123';
SET s3_use_ssl=false;
SELECT * FROM 's3://my-bucket/my-file.snappy.parquet';
```

jwills: gotcha, cool-- I'm thinking, looking at https://duckdb.org/docs/sql/configuration, that I really should just generalize these settings into like a `duckdb_config` entry in the creds and let the user config whatever they want up in here.

edgarrmondragon: That makes sense. Want me to take a stab at that and add deprecation warnings for the existing settings?

jwills: Ah I just threw this together over here before I saw your comment: https://github.com/jwills/dbt-duckdb/pull/25

jwills: The deprecation approach is much nicer but I'm not too-too worried about it; the migration from the old config to the new is pretty lightweight.

edgarrmondragon: > the migration from the old config to the new is pretty lightweight.

fair enough 👍

edgarrmondragon:
Superseded by #25

PR #25, opened 2022-10-15
Generalize the extensions/settings config for DuckDB and update the README

### LINKED ISSUE #23: Support for `s3_endpoint`

jwills:
Instead of special-casing the s3 configuration, let's simply add a `settings` dictionary to the profile where a user can specify any/all configuration settings for their DuckDB connection that they might need; fixes #23 in a more general way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,10 +22,29 @@ There is also a `database` field defined in the `DuckDBCredentials` class for co
 but it defaults to `main` and setting it to be something else will likely cause strange things to happen that I cannot fully predict,
 so, ya know, don't do that.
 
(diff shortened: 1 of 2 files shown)


jwills:
@edgarrmondragon does something like this work for you? Any feedback?

edgarrmondragon:
> @edgarrmondragon does something like this work for you? Any feedback?

@jwills no feedback other than this looks good 😄.

A good follow-on might be a PR for the dbt reference: https://github.com/dbt-labs/docs.getdbt.com/blob/current/website/docs/reference/warehouse-setups/duckdb-setup.md

PR #28, opened 2022-10-20
feat: python model

### LINKED ISSUE #27: Python models
It would be great to have a support for dbt Python Models. My use case is primarily to read/write into Delta tables (hope it will be obsolete with [duckdb-deltatable-extension](https://github.com/Mause/duckdb-deltatable-extension)), but there are many other possibilities.

I have already created a prototype, but it is an ugly duckling that supports only single thread (which is okish for my use-case). But I do not expect PR to be merged anytime soon.

P.S. I have created this issue because it is always nice to have both PR and issue :)

tomsej:
closes #27

Todo:
- [x] missing support for multithreading
- [x] incremental model

### DIFF
--- a/dbt/adapters/duckdb/impl.py
+++ b/dbt/adapters/duckdb/impl.py
@@ -1,14 +1,52 @@
-from dbt.adapters.duckdb import DuckDBConnectionManager
+from dbt.adapters.duckdb.connections import DuckDBConnectionManager
 from dbt.adapters.sql import SQLAdapter
+from dbt.contracts.connection import AdapterResponse
+from dbt.exceptions import InternalException, RuntimeException
 
 
 class DuckDBAdapter(SQLAdapter):
     ConnectionManager = DuckDBConnectionManager
 
     @classmethod
-    def date_function(cls):
+    def date_function(cls) -> str:
         return "now()"
 
     @classmethod
-    def is_cancelable(cls):
+    def is_cancelable(cls) -> bool:
         return False
+
+    def valid_incremental_strategies(self):
+        """DuckDB does not currently support MERGE statement."""
+        return ["append", "delete+insert"]
+
+    def commit_if_has_connection(self) -> None:
+        """This is just a quick-fix. Python models do not execute begin function so the transaction_open is always false."""
+        try:
+            self.connections.commit_if_has_connection()
+        except InternalException:
+            pass
+
+    def submit_python_job(self, parsed_model: dict, compiled_code: str) -> AdapterResponse:
+
+        connection = self.connections.get_if_exists()
+        if not connection:
+            connection = self.connections.get_thread_connection()
+        con = connection.handle._conn
+
+        def load_df_function(table_name: str):
+            """
+            Currently con.table method dos not support fully qualified name - https://github.com/duckdb/duckdb/issues/5038
+
+            Can be replaced by con.table, after it is fixed.
+            """
+            return con.query(f"select * from {table_name}")
+
+        try:
+            exec(compiled_code, {}, {"load_df_function": load_df_function, "con": con})
+        except SyntaxError as err:
+            raise RuntimeException(
+                f"Python model has a syntactic error at line {err.lineno}:\n" f"{err}\n"
+            )
+        except Exception as err:
+            raise RuntimeException(f"Python model failed:\n" f"{err}")
+        return AdapterResponse(_message="OK")
--- a/dbt/include/duckdb/macros/adapters.sql
+++ b/dbt/include/duckdb/macros/adapters.sql
@@ -28,7 +28,7 @@
   {{ return(run_query(sql)) }}
 {% endmacro %}
 
-{% macro duckdb__create_table_as(temporary, relation, sql) -%}
+{% macro duckdb__get_create_table_as_sql(temporary, relation, sql, language='sql') -%}
   {%- set sql_header = config.get('sql_header', none) -%}
 
   {{ sql_header if sql_header is not none }}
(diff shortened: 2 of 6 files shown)


tomsej:
Yes, exactly! Python model originally should work as independent process to the main flow. But I think I have already found the solution for this. Just bear with me.

jwills:
Other question: do we actually need to write these materializations  (table, incremental) ourselves in dbt-duckdb? Or is that just a placeholder b/c we haven't upgraded to dbt-core 1.3.0 yet?

tomsej:
This is an excellent question. Currently, we need to have those materializations since default ones in `dbt-core` do not support Python models. Materializations were changed a lot recently, but all adapters with Python models (Snowflake, BQ, Spark) are still using their own (based on old structure), that differ a lot to each other. What I did is something in the middle, I used the new version of the templates and added support for python models. 
**tldr;** it is a mess now and most likely it will be the matter of change. I will ask devs on dbt Slack.

About 1.3.0 we have already implicitly upgraded. You can verify it if you currently install `dbt-duckdb` in the fresh venv. The reason is that `dbt-core` version is not pinned here - https://github.com/jwills/dbt-duckdb/blob/bcf33d845587f507aa99a1980f69822d0f3dcf30/setup.py#L47. It says install any version equal or bigger than 1.2.0. It is tricky. Think you should use `~=` operator that would install any minor version bigger than 1.2.0.

jwills:
Yep, that minor version thing should be fixed-- thanks!

tomsej:
Yeah I am doing it in #29

tomsej:
Think I found the solution for both multithreading and incremental models. I just need to add some more tests.

dbt/adapters/duckdb/impl.py
```diff
@@ -1,14 +1,52 @@
...
+        if not connection:
+            connection = self.connections.get_thread_connection()
+        con = connection.handle._conn
+
+        def load_df_function(table_name: str):
+            """Currently con.table method dos not support fully qualified name - https://github.com/duckdb/duckdb/issues/5038"""
+            return con.query(f"select * from {table_name}")
+
+        try:
+            exec(compiled_code)
```
jwills, line 45: FYI that this approach very much resonates with me, but I sort of suspect that we should do some work to limit the scope of the `exec` call by specifying the `globals` and `locals` dictionaries, right? Like we would e.g. not want the script running to be able to access/modify the state of the dbt run beyond whatever scope we passed into it.

tomsej: Honestly, I was not thinking about it at all :). It is a really great point. I have limited both dictionaries to a bare minimum. `load_df_function` can be deleted in the future after https://github.com/duckdb/duckdb/issues/5038 is fixed.

tomsej:
Think now it should work fine. Incremental model is supporting only `append` and `delete+insert` since DuckDB is not supporting the `merge` statement. Any idea if I forgot something @jwills? Thank you!

ChenyuLInx:
Hey @tomsej @jwills, sorry I am late to the game, this looks great!

> Currently, we need to have those materializations since default ones in dbt-core do not support Python models

I agree, should be a small lift to actually do this in `dbt-core` so other adapters that just use the default materializations macros don't have to do similar thing again. I will create a branch in core and test it out.

Out of all the adapters dbtLabs maintain, the `get_create_table_as_sql` macro seems only exists in core and not used in other places. It looks quite messy in terms of which macros should be used there. I will check with Jerco to see if he has any thoughts on which one to settle on in the longer term. For python model in other adapters, we overwritten the macro `create_table_as` and added a new parameter `language`. I think with some updated in dbt-core we should be able to do the same thing for duckdb. I will keep you all posted.

tomsej:
Thank you, @ChenyuLInx, for reaching out. I really appreciate it.

> I agree, should be a small lift to actually do this in `dbt-core` so other adapters that just use the default materializations macros don't have to do similar thing again. I will create a branch in core and test it out.

That would be really cool. Please let me know when it will be ready. Thank you!

> For python model in other adapters, we overwritten the macro `create_table_as` and added a new parameter `language`. I think with some updated in dbt-core we should be able to do the same thing for duckdb. I will keep you all posted.

In the last commit, I replaced `get_create_table_as_sql` and `get_create_table_as_python` macros with the `create_table_as` statement with an extra `language` parameter. So, the materialization macros contain only minor changes now. If materializations support the python models in the `dbt-core`, we will be able to delete the templates from this repo. I don't think this should currently be a blocker for the new release of `dbt-duckdb`.

ChenyuLInx:
> That would be really cool. Please let me know when it will be ready. Thank you!

@tomsej since we just released dbt-core 1.3, the earliest time it can get in is 1.3.1, which doesn't have a planed date yet I think. Totally agree it shouldn't be a blocker and what you got right now seems like the way to go for dbt-duckdb 1.3 IMHO. And we can follow up again to remove those materialization macros when dbt-core 1.3.1 got released. #releaselogistics

dbt/adapters/duckdb/impl.py
```diff
@@ -1,18 +1,52 @@
...
+        def load_df_function(table_name: str):
+            """
+            Currently con.table method dos not support fully qualified name - https://github.com/duckdb/duckdb/issues/5038
+
+            Can be replaced by con.table, after it is fixed.
+            """
+            return con.query(f"select * from {table_name}")
+
+        try:
+            exec(compiled_code, {}, {"load_df_function": load_df_function, "con": con})
```
turbo1912, line 45: great job @tomsej! a completely passer by comment, so feel free to ignore it. 

One piece of feedback we got from dbt labs folks was that they didn't want python models to run in the same process where dbt-core is running, for good reasons.. Looks like @ChenyuLInx had a look at this PR and was not worried about this so maybe my information is outdated.

jwills: @turbo1912 I think given that DuckDB itself is running in the same process as dbt, that ship has sailed; dbt-duckdb is the 🏴‍☠️ dbt adapter. 😉

tomsej: Yohoho, @turbo1912 :). Think it would be a problem if this adapter would be supported in cloud, but I do not think it will ever be.

tomsej: We also need to run it in the same process, since you can have only one connection to DuckDB

Issue #26, opened 2022-10-19
Give a duck some ❤️


tomsej:
Ducks are fantastic animals but need some care, just like a code. [Here](https://github.com/dbt-labs/dbt-database-adapter-scaffold) is a repository for scaffolding dbt adapter. Unfortunately, this awesome project does not comply to that, which makes contribution more challenging than it has to be.

Would it be okay if I brought the structure closer to the template?

jwills:
Hey @tomsej! I haven't seen that scaffolding repo before-- where is it used? And how does it work?

tomsej:
Think the repo is used in all the main adapter repositories like [dbt-spark](https://github.com/dbt-labs/dbt-spark) or [dbt-snowflake](https://github.com/dbt-labs/dbt-snowflake). It is a [cookiecutter](https://github.com/cookiecutter/cookiecutter) template. So at the beginning of the development you just run `cookiecutter gh:dbt-labs/dbt-database-adapter-scaffold` and it scaffolds project structure for you. But now, when project is already up, it would be better just to copy some content.

There are some particular paint points that are solved in cookiecutter:
- missing `dev-requirements.txt`, so you need to find required dev dependencies yourself.
- missing tox configuration and github actions configuration. So testing on multiple python version is harder.
- semantic versioning of the package.

jwills:
Ah, gotcha-- I would welcome the changes needed to support this via a PR, thanks!

PR #29, opened 2022-10-20
feat: adapt to scaffolding template

tomsej:
This is a first iteration on closing(#26). I also included .devcontainer for easier deployment. Also I changed some code that `mypy` was complaining about. But most of the changes are caused by `black` formatter.


jwills (commented):
This all looks great, thank you so much @tomsej! I'm a little iffy on the devcontainer stuff as I do not use them, but if it would make your life significantly better to include it, I'm happy to do so as a way of saying "thank you" :)

tomsej:
Please excuse the delayed response! It is not you who should thank, it should be me. I know pretty well how annoying it is to have contributors 😄.

Devcontainers are wonderful, especially for lazy guys like me.

And big thanks for merging that!!

PR #31, opened 2022-10-21
Ensure that we always keep a connection open when we're running dbt-duckdb in memory

### LINKED ISSUE #30: dbt models cannot see each other when using the :memory: database.
## Describe the bug
dbt models cannot see each other when using the :memory: database.

## To Reproduce

1. create two models. table_one.sql & table_two.sql
2. table_one.sql should be "SELECT 'test' AS value".
3. table_two.sql should be "SELECT * FROM {{ ref( 'table_one' ) }}
4. table_two will fail to build with following error: "Catalog Error: Table with name table_one does not exist"

## Expected behavior
dbt build should succeed.

## Your environment
Version of dbt-duckdb: ~=1.2.0

jwills:
Fixes #30

### DIFF
--- a/dbt/adapters/duckdb/connections.py
+++ b/dbt/adapters/duckdb/connections.py
@@ -106,13 +106,16 @@ class DuckDBConnectionManager(SQLConnectionManager):
                         for extension in credentials.extensions:
                             cls.CONN.execute(f"INSTALL '{extension}'")
 
-                connection.handle = DuckDBConnectionWrapper(cls.CONN.cursor(), credentials)
+                connection.handle = DuckDBConnectionWrapper(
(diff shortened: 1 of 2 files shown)


jwills:
@tomsej here's what I'm thinking: I want to merge this bug fix PR, cut a release of dbt-duckdb `1.2.3` that includes this and the new `settings` config stuff from #25, create a `1.2-release` branch so we have a place to do any other bug fixes on the `1.2.x` line that are required, and then we can devote the `master` branch to merging in the 1.3.0 feature work you have in #28 and #32-- does that work for you? /cc @matsonj

tomsej:
Yes, this makes a lot of sense! Thank you so much, @jwills!

Issue #19, opened 2022-10-09
Parquet materialization 


jpmmcneill:
👋

Hey!

Something I've been thinking about is using parquet as a _materialization_. To be specific, one where `dbt-duckdb` would use the underlying duckdb connection as a shim to edit parquet rather than adding some tables into the `.duckdb` file.

I'm not sure if it's possible to override what `ref` does in an adapter, but this would roughly need to do two things:

1. Make a new materialization that just took the query for some model and executes
`copy (select ...) to 'model_name.parquet' (format parquet)` rather than`create table model_name as (select ...)` for parquet materialized models.
2. Update ref to use `read_parquet(ref'd model name)` rather than schema.table_name _if the model being referenced was materialized as parquet_.

Would love to hear what other people think / if this would be useful! 

This is loosely related to: https://github.com/jwills/dbt-duckdb/issues/15

jpmmcneill:
It goes without saying that I'm happy to take a stab at it, but i'd love to get a temperature check from folks first 😄

jwills:
Both of those things seem doable to me, but there are a few dragons we need to be mindful of. The big one is handling backups: when dbt materialized tables/views, it creates backups of the old relations via `rename` operations before it actually materializes the new one, so that if the new materialization fails for some reason, we haven't destroyed the old relation and can recover its contents. I just did some local testing of this idea with some `COPY (SELECT ...) TO 'output.parquet'` and it seemed like things mostly worked (i.e., various kinds of query failures did _not_ seem to blow away the file until the query succeeded), but I'd want to be sure to test this out pretty thoroughly before recommending it to folks as a more general pattern.

Second, it seems like a pretty easy lift to do this for CSV files at the same time, right? You would need a way for the `config(...)` macro block to accept options related to e.g. compression codecs for either format, and then of course [CSVs have a bunch more parameters](https://duckdb.org/docs/data/csv) you can config, but it seems like it's the same idea and would probably be generally useful as well.

Yeah, with those two things in mind, I think you could build something off of the changes in this branch to introduce a `DuckDBRelation`: https://github.com/jwills/dbt-duckdb/compare/master...jwills_file_based_dbs where you could be mindful of what your target materialization was and implement the `render` method accordingly.

[...]

Issue #15, opened 2022-09-27
dbt models that are parquet files


sryza:
I haven't thought through this deeply. It might not make sense, it might require changes to dbt, or it might already work? But I wanted to raise it just in case, because it would help me out with something I'm building.

Could we enable configuring dbt-duckdb such that 
```
select
    bla
from {{ ref("orders") }}
```

compiles to

```
select
    bla
from 's3://bucket/orders.parquet'
```
?

jwills:
Yeah, I'm pretty sure that works already! You need to specify the file location as a `source` and use the filepath as an `identifier` ala https://docs.getdbt.com/reference/resource-properties/identifier

I suspect I could make the process for wiring this up a little bit nicer, just LMK how it works for you!

jwills:
Oh actually it might be 1) a `source` with an `identifier` that looks like `"read_parquet('s3://bucket/key')"` modulo https://duckdb.org/docs/guides/import/s3_import

(and of course you would also have to have the AWS S3 creds wired up in the credentials)

sryza:
If I understand correctly, that means dbt sources can be parquet files. Would that approach extend to models?

I.e., in my dream imagination of how this works, when you run a model, it creates a parquet file in S3. If I understand how DuckDB works, this might need to be a two-stage process, where one SQL statement creates the table in DuckDB and then a second one copies it to Parquet. Which might be a little too gnarly?

The advantage of this would be making it easier to use DuckDB in a distributed setting: S3 becomes the coordination point and the database file becomes ephemeral.

jwills:
So you're saying in this execution model that I could fire up multiple DuckDB processes and farm the queries against the Parquet data in S3 (or wherever-- a shared volume could work as well) out to them? That is interesting.

sryza:
Exactly!

sgoley:
Honestly, sounds a lot like what you are looking for @sryza is [dbt-external-tables](https://github.com/dbt-labs/dbt-external-tables) to be have an adapter for duckdb? 

As far as the development of that goes, it looks like it couldn't be purely "external" in the way that the others adapters for Snowflake or Bigquery operate, but you might be able to make something work through the [COPY](https://duckdb.org/docs/sql/statements/copy) api since that supports external load of parquet files from a remote storage.

That would be my starting point.

jwills:
I started experimenting with this idea over here: https://github.com/jwills/dbt-duckdb/compare/jwills_file_based_dbs?expand=1

[...]

PR #33, opened 2022-10-25
feat: add support for external tables

tomsej:
Add new type of materialization - `external_table`. It allows to dump data to `parquet` or `csv` and dbt creates a view on top.

#19 #15


jwills:
Hey @tomsej just wanted to call out that dbt ships with a built-in `RelationType` called `external`: https://github.com/dbt-labs/dbt-core/blob/c2856017a12b947f9d61901b187bfab177516579/core/dbt/contracts/relation.py#L21

which I think is used as a part of https://github.com/dbt-labs/dbt-external-tables

...which could potentially be a useful reference for this impl? At the very least, I would want to minimize confusion between the dbt-duckdb `external_table` and the dbt-external-tables `external` relation type.

jwills:
Looked at that dbt-external-tables repo some more which seems, uhh, not actively developed; the right move might simply be to use the `RelationType` w/o actually using the package itself and make it our own

tomsej:
Oh, that is a good point. I am awful at naming the things 😕. What are your thoughts on calling it a `file` materialization? Or do you want to just use an `external`?

jwills:
If it was up to me I would simply call it `external` and make it our own, up-to-and-including stealing any macros we needed to support it from that repo. ;-)

jwills:
(FWIW I am also awful at naming things, but I make up for it by being really awesome at cache invalidation)

tomsej:
Stealing?! I like your style! 

`external` works for me too!

jwills:
good artists copy, great artists steal, etc.

tomsej:
Sorry, I can't get it out of my head, what's next in that etc - awesome artists rape?

PR #34, opened 2022-10-25
feat: csv and parquet materialization

jpmmcneill:


dbt/adapters/duckdb/relation.py
```diff
@@ -0,0 +1,45 @@
...
+
+class DuckDBRelationType(StrEnum):
+    Table = "table"
+    View = "view"
+    CTE = "cte"
+    Parquet = "parquet"
+    CSV = "csv"
+
+@dataclass(frozen=True, eq=False, repr=False)
+class DuckDBRelation(BaseRelation):
```
jpmmcneill, line 19: had to add this class back in because of the type changes + render overriding

jpmmcneill:
this is currently based on a branch that isn't master, hence the gnarly commit history. I'll overwrite this tomorrow I think!

PR #35, opened 2022-10-27
feat: add support for external materialization

tomsej:


dbt/adapters/duckdb/impl.py
```diff
@@ -2,6 +2,8 @@
 from dbt.adapters.sql import SQLAdapter
 from dbt.contracts.connection import AdapterResponse
 from dbt.exceptions import InternalException, RuntimeException
+from typing import Sequence
+from dbt.adapters.base.meta import available
```
jwills, line 6: import ordering issue here-- I thought we had linting for this now tho?

dbt/adapters/duckdb/impl.py
```diff
@@ -15,7 +17,19 @@ def date_function(cls) -> str:
     def is_cancelable(cls) -> bool:
         return False
 
-    def valid_incremental_strategies(self):
+    @available
+    def location_exists(self, location: str) -> bool:
+        try:
+            self.execute(
+                f"select 1 from '{location}' where 1=0",
```
jwills, line 24: ah yay this is very clever!

dbt/include/duckdb/macros/materializations/external.sql
```diff
@@ -0,0 +1,91 @@
+{% materialization external, adapter="duckdb", supported_languages=['sql', 'python'] %}
```
jwills, line 1: does this really work for python too?

dbt/include/duckdb/macros/materializations/external.sql
```diff
@@ -0,0 +1,91 @@
...
+
+  {%- set location = config.require('location') -%}
+  {%- set format = config.get('format', default='parquet') -%}
+  {%- set delimiter = config.get('delimiter', default=';') -%}
+  -- set language - python or sql
+  {%- set language = model['language'] -%}
+
+  {%- set target_relation = this.incorporate(type='view') %}
+  -- check if the location exists
+  {%- set location_exists = location_exists(location) %}
```
jwills, line 11: This can be:

`{%- set location_exists = adapter.location_exists(location) %}`

unless you had a reason for adding in extra macro as a layer of indirection?

dbt/adapters/duckdb/impl.py
tomsej: Yeah, we do. I had some problems with my Mac, so I used the other one. And did not install pre-commit. My bad

dbt/adapters/duckdb/impl.py
tomsej: Thanks

dbt/include/duckdb/macros/materializations/external.sql
tomsej: Yep, it supports Python too

dbt/include/duckdb/macros/materializations/external.sql
tomsej: We need to get rid of the database name, so that is why

dbt/adapters/duckdb/impl.py
tomsej: Sorry, my bad, I thought that scaffolding template has import ordering in pre-commit config. Seems it is not true. So I reformated all files and added check into pre-commit. It is already merged into `master`

tomsej:
@jwills I added tests too. So if you are ok with my answers to your comments, we can merge.

jwills:
@tomsej ack, I just merged in @AlexanderVR's concurrency test fix that caused a conflict for you, sorry about that! Otherwise I am good to go, let's merge this and prep a new release!

tomsej:
NP, @jwills! I fixed that and merged this PR.

PR #37, opened 2022-10-31
fix: development env requirements

AlexanderVR:
- bump mypy version in dev-requirements to allow install on ARM-based macs. 
- bump black version in pre-commit to fix override of default_language_version. ref: https://github.com/psf/black/pull/2430


tomsej:
@AlexanderVR Thanks, I just copied the versions from `dbt-adapter-scaffold`. I have also added dependabot workflow to check versions in the futere.

PR #38, opened 2022-10-31
fix: concurrency tests

AlexanderVR:
Patch the COLUMNS_EQUAL_SQL template -- used in the adapter method `get_rows_different_sql` -- to not use the same identifiers as the tested relations in the TestConcurrency test cases. This avoids a duckdb binding issue and allows the original test cases to be evaluated.


jwills:
LGTM @AlexanderVR -- thank you!

PR #46, opened 2022-11-01
Minimal amount of effort required to make external locations work for sources

jwills:
I'm taking advantage of the fact that the `external` construct is already defined on the `ParsedSourceDefinition` in dbt-core to make this work; this is extremely trivial right now as implemented but you could imagine where we could make it much richer if we were so inclined.

https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/contracts/graph/parsed.py#L637

Putting this up now to get some eyes on it while I work on the tests


tomsej:
Looks great! Like the simplicity

jwills:
@tomsej my python-fu is not strong enough to understand what mypy is unhappy about ^^

tomsej:
I will take a look tomorrow.

tomsej:
I wrote you the comment to the part that makes the problem + you should run `pre-commit run --all-files`. In your IDE you are probably using some kind of import sorter. The most used in `pre-commit` is this one https://github.com/asottile/reorder_python_imports that is modifying this:

```from os.path import abspath, exists```

to

```
from os.path import abspath
from os.path import exists
```
It is gettin more popular lately since it helps to prevent the merging conflict. But if you dislike it, we can use a different one.

**Review by tomsej — CHANGES_REQUESTED** [review]

File: dbt/adapters/duckdb/relation.py [hunk 1]
```diff
@@ -0,0 +1,25 @@
...
+class DuckDBRelation(BaseRelation):
+    external_location: Optional[str] = None
+
+    @classmethod
+    def create_from_source(cls: Type[Self], source: ParsedSourceDefinition, **kwargs: Any) -> Self:
+        if source.external:
+            # TODO(jwills): this is trivial and could be made much richer, just want to know
+            # if it works
+            kwargs["external_location"] = f"'{source.external.location}'"
+        return super().create_from_source(source, **kwargs)
```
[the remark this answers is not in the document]
jwills: Ah yeah so this is my ignorance of the details of the impl of python classmethods-- how exactly do I signal that I want the call to go to the parent class' classmethod instead of this one? It's not, `super()`, it's `cls`? Really?

jwills:
@tomsej other questions:

1) Advice on how to test this under the `functional` framework? Did you just go find similar tests in the upstream dbt project and copy them over w/modifications?
2) Is it okay if I take a crack at the top-level `external_root` stuff for the `external` materializations we create? I imagine that at some point in the future I'll want to create some similar sort of config option for sources so I'd like to keep the two sets of config options in sync with one another.

dbt/adapters/duckdb/relation.py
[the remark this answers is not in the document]
tomsej: Oh no, it is my bad I did not realize you are calling the same method name. I guess it should be `super(DuckDBRelation, cls)`, will take a look if mypy is ok with it.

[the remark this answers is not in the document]
tomsej: Oh does not help. But it seems like a known issue - https://github.com/python/mypy/issues/12885. So we can probably ignore:

```
return super().create_from_source(source, **kwargs)  # type: ignore
```

dbt/adapters/duckdb/relation.py
```diff
@@ -0,0 +1,43 @@
...
+        # Some special handling here to allow sources that are external files to be specified
+        # via a `external_location` meta field. If the source's meta field is used, we include
+        # some logic to allow basic templating of the external location based on the individual
+        # name or identifier for the table itself to cut down on boilerplate.
+        ext_location = None
+        if "external_location" in source.meta:
+            ext_location = source.meta["external_location"]
+        elif "external_location" in source.source_meta:
+            # Use str.format here to allow for some basic templating outside of Jinja
+            ext_location = source.source_meta["external_location"].format(
```
tomsej, line 27: Should not we allow templating in both `source.meta` and `source.source_meta`?

dbt/adapters/duckdb/relation.py
```diff
@@ -0,0 +1,43 @@
...
+                name=source.name,
+                identifier=source.identifier,
+            )
+        if ext_location:
+            # If it's a function call or already has single quotes, don't add them
+            if "(" not in ext_location and not ext_location.startswith("'"):
```
tomsej, line 33: Hacky, but I do not know how to do it better :/

jwills: Yah, agreed

dbt/adapters/duckdb/relation.py
jwills: I started that direction at first, but then I wasn't sure; if you're specifying a location in `source.meta`, then you already know the name of the table, so I wasn't sure how the templating helped-- are you thinking it's just better for consistency purposes, even though it's not as helpful when it's on `source.meta`?

tomsej: I was more thinking about consistency. But I guess this makes sense too

tomsej: Or more precisely that source.meta is setting common for all underlying sources.

jwills: So the naming is confusing here-- the `source` refers to the node (aka the table/view/file) itself, which is allowed to have its own `meta` dictionary, as well as the top-level `sources:` object in the yaml (which includes a list of `tables`) that is _also_ allowed to have its own `meta` dictionary.

But leaving all of that aside, your point about consistency is a good one b/c there's no harm in enforcing it here, so I will make a change to that effect. I'm also going to add in some tests I derive from https://github.com/dbt-labs/dbt-core/tree/main/tests/functional/sources and will then merge it and, dare I say it, cut a release?!

jwills: Nothing else you want to get in to 1.3.0, right?

PR #47, opened 2022-11-02
feat: add support for glue catalog

tomsej:
Experimental support for registering external tables in Glue Catalog.


tomsej:
For some reason tests pass if i run `pytest tests/unit/test_glue.py` but fails if I run `pytest tests/unit/`. Will take a look

PR #49, opened 2022-11-02
feat: `external_root` profile setting as dir to write external tables when no location is specified

jwills:
Per a discussion on #19, I'm adding an `external_root` to the top-level `duckdb` profile object (specified via the `Credentials` class) which defaults to `.` (the current working directory) but can be overridden to make it easy to specify a directory or S3 bucket where the parquet/csv files for any `external` materialization should be written to in the form of `{{ external_root }}/{{ identifier }}.{{ format }}` per the `external_location` macro.

### DIFF
--- a/dbt/adapters/duckdb/connections.py
+++ b/dbt/adapters/duckdb/connections.py
@@ -34,6 +34,10 @@ class DuckDBCredentials(Credentials):
     # (and extensions may add their own pragmas as well)
     settings: Optional[Dict[str, Any]] = None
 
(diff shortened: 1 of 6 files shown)


tomsej (approved):
Looks good to me!

tomsej:
[okay finally have my black settings right](https://github.com/jwills/dbt-duckdb/pull/49/commits/82babcebbc42f5414d72473b734cfcdc6c9a850e) - love this commit message 😄

PR #56, opened 2022-11-12
Add render support for all of the string config options for external materializations

### LINKED ISSUE #55: External materialization file paths do not support jinja references
Consider the following config block:

```
{{ 
    config(
        materialized='external', 
        location='/path/to/folder/{{ this.name }}.parquet'
) }}
```

This should write the model name to the folder specified.

However, what happens is the name is written literally as ```{{ this.name }}.parquet``` as seen in the screenshot below.

705B6873-6F32-4976-BDA0-8F527E0E0FE6

I would expect that the external materialization works with the same syntax as this (but it does not):

```
COPY (SELECT * FROM {{ this }} ) TO '/tmp/data_catalog/conformed/{{ this.name }}.parquet' (FORMAT 'parquet')
```

jwills:
Fixes #55 

Obvi we have a workaround that is fine, but I think it will be more intuitive for folks if the `location` arg (and any other string options for `external`) work just like `pre_hook` and `post_hook` and allow configuration via jinja template args.

### DIFF
--- a/dbt/include/duckdb/macros/materializations/external.sql
+++ b/dbt/include/duckdb/macros/materializations/external.sql
@@ -1,10 +1,10 @@
 {% materialization external, adapter="duckdb", supported_languages=['sql', 'python'] %}
 
-  {%- set format = config.get('format', default='parquet') -%}
-  {%- set location = config.get('location', default=external_location(format)) -%}
-  {%- set delimiter = config.get('delimiter', default=',') -%}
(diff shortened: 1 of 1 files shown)


tomsej:
Good idea @jwills! So you can use eg. `this.identifier` in the location?

jwills:
> Good idea @jwills! So you can use eg. `this.identifier` in the location?

exactly; `location="/here/is/some/path/to/{{ this.identifier }}.parquet"` will work seamlessly

PR #68, opened 2022-11-29
Use an importlib-based approach to loading and executing dbt Python models

### LINKED ISSUE #63: Python model fails with import statement outside of function
Python models aren't working as I expect. This basic example fails:

```py
import pandas as pd

def model(dbt, session):

    return pd.Series(range(10)).to_frame()
```

```
15:32:53  Runtime Error in model 03 (days/03.py)
15:32:53    Python model failed:
15:32:53    name 'pd' is not defined
```

However, simply moving the import statement into the function allows it to work.
```py
def model(dbt, session):
    import pandas as pd
    return pd.Series(range(10)).to_frame()
```

Versions:
- dbt-duckdb 1.3.2
- dbt 1.3.1
- Python 3.10.7

jwills:
Fixes #63 

This ended up being kind of neat and I think better than the `exec` based approach: the idea here is that we treat each dbt Python model as its own Python module that we dynamically load and execute using the methods inside of `importlib.util`. Once that is done, we can call the `dbtObj`, `model`, and `materialize` functions that we declared inside of the `.py` file as if they were simply functions on the module, so we get:

a) Working top-level imports (to fix #63)
b) Isolation of the code in each model file in a way that we don't pollute the globals

...which is pretty cool!

### DIFF
--- a/dbt/adapters/duckdb/impl.py
+++ b/dbt/adapters/duckdb/impl.py
@@ -1,3 +1,4 @@
+import os
 import importlib.util
 import tempfile
 from typing import List
@@ -104,7 +105,7 @@ class DuckDBAdapter(SQLAdapter):
         except Exception as err:
             raise RuntimeException(f"Python model failed:\n" f"{err}")
         finally:
-            mod_file.close()
+            os.unlink(mod_file.name)
         return AdapterResponse(_message="OK")
 
     def get_rows_different_sql(
--- a/dbt/adapters/duckdb/impl.py
+++ b/dbt/adapters/duckdb/impl.py
@@ -1,3 +1,5 @@
+import importlib.util
+import tempfile
 from typing import List
 from typing import Optional
 from typing import Sequence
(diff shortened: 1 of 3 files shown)


tomsej:
Super interesting @jwills! Honestly, I was thinking the same when I wrote `exec` approach and also was considering to rewrite it! Great job!!!!

dbt/adapters/duckdb/impl.py
```diff
@@ -86,14 +89,32 @@ def load_df_function(table_name: str):
             """
             return con.query(f"select * from {table_name}")
 
+        identifier = parsed_model["alias"]
+        mod_file = tempfile.NamedTemporaryFile(suffix=".py", delete=False)
+        mod_file.write(compiled_code.lstrip().encode("utf-8"))
+        mod_file.close()
```
tomsej, line 95: Does not close a temp file delete the file?

tomsej: Ohh I see. You set delete to False

PR #70, opened 2022-12-11
Use better, non-regex based syntax for handling the limit case inside of the listagg macro

jwills:
Learned this trick thanks to the DuckDB discord; replaces my very hacky, regex-based approach to handling the `limit` case correctly.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,6 +49,8 @@ jobs:
 
       - name: Set up Python
         uses: actions/setup-python@v4
+        with:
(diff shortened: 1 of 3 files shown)


jwills:
@tomsej when you have a moment (not urgent) can you look at why the python stuff is failing above? Like, nothing changed AFAICT but the Actions seem like they can no longer find Python.

tomsej:
Hi @jwills , sorry I had covid recently, than kids and wife, so have no capacity to do anything else. Will take a look tomorrow

PR #72, opened 2022-12-12
ext_location: str.format with schema

keiththompson:
I've been using this recently and found myself putting some vars in dbt_project.yml when I think just using the schema name similar to name or identifier probably makes more sense. 

Let me know if there is a different approach I should be taking here! I'm certainly more used to consuming dbt than contributing to adapters etc.


jwills:
@keiththompson this looks great, appreciate you trying it out and adding in some data that proved useful to you! I was going back-and-forth on whether I should include the `schema` in the format params so I am happy to bring it in here.

PR #77, opened 2023-01-10
Fix python models not inheriting AWS credentials

AlexanderVR:
Solution: use the cursor which has the credentials set, instead of the raw connection.


jwills (approved):
Ack! Thank you so much!

Issue #79, opened 2023-01-18
Models that depend on external models do not automatically register their dependencies as views


AlexanderVR:
Suppose we specify `+materialization: external` in our dbt_project.yml.

Initially we run all dbt computations using an in-memory db, materializing all the models as parquet files.

Subsequently we change one dbt model A, and want to recompute model A (and possibly downstream dependencies as well).

Now, if A depends on upstream models B and C, trying to run just A will result in a failure. Because, B and C are not registered as views yet (in-memory db!), despite all tables being specified as external. This forces the user to always recompute upstream dependencies just to trigger the registration logic https://github.com/jwills/dbt-duckdb/blob/master/dbt/include/duckdb/macros/materializations/external.sql#L51-L55

I'd like to propose putting in a check in the materialization macros to 
- loop through the parent nodes
- for each parent_node, if load_cached_relation(parent_node) is None, call the registration logic.

Does this sound reasonable? Happy to send a PR.

jwills:
The problem as you describe it makes sense, but I still don't totally grok the use case where folks are running stuff entirely in-memory over-and-over again without using a DuckDB database file, even if it's just to keep track of view definitions and how they map to external parquet files. Where does this sort of thing come up?

AlexanderVR:
Suppose you have a `dbt` graph containing several hundred models. Different sets of sources are updated at different times (or one might update a few model definitions), and you spin off Kubernetes jobs that recompute the affected subsets of the graph.

With a centralized DuckDB database file, you would need some locking mechanism allowing only one job at a time to take ownership of the DuckDB file. 

There is also the temporary instability of the DuckDB storage format, which makes sharing the db files across different jobs with possibly different DuckDB versions a bit annoying.

jwills:
yes, got it-- thank you so much for taking the time to explain it! A PR would be most welcome!

PR #80, opened 2023-01-21
Register views for upstream external models.

AlexanderVR:
Closes https://github.com/jwills/dbt-duckdb/issues/79

This requires a (minor) breaking change:
* The `external_location(format)` macro is now `external_location(this, format)`.

If one is overriding this macro in their dbt project, they will just have to convert their macro to the new signature.

This change is needed in order to determine the external location of upstream models, not just the current model. We explicitly shadow `this` to prevent confusion between the current model and the upstream model when specifying the location.

### DIFF
--- a/dbt/include/duckdb/macros/utils/external_location.sql
+++ b/dbt/include/duckdb/macros/utils/external_location.sql
@@ -1,3 +1,3 @@
-{%- macro external_location(format) -%}
-  {{- adapter.external_root() }}/{{ this.identifier }}.{{ format }}
+{%- macro external_location(relation, format) -%}
+  {{- adapter.external_root() }}/{{ relation.identifier }}.{{ format }}
 {%- endmacro -%}
--- a/dbt/include/duckdb/macros/utils/upstream.sql
+++ b/dbt/include/duckdb/macros/utils/upstream.sql
@@ -1,0 +1,25 @@
+{%- macro register_external_upstream() -%}
+{% if execute %}
+  {% for node in model['depends_on']['nodes'] %}
+    {% set upstream = graph.nodes.get(node) %}
+    {% if upstream %}
+      {%- set upstream_rel = api.Relation.create(
+        database=upstream['database'],
+        schema=upstream['schema'],
+        identifier=upstream['alias']
+      ) -%}
+      {%- if upstream.config.materialized=='external' and not load_cached_relation(upstream_rel) -%}
+        {%- set format = render(upstream.config.get('format', 'parquet')) -%}
+        {%- set upstream_location = render(
+            upstream.config.get('location', external_location(upstream_rel, format)))
+        -%}
+        {% call statement('main', language='sql') -%}
+          create or replace view {{ upstream_rel.include(database=False) }} as (
+            select * from '{{ upstream_location }}'
+          );
+        {%- endcall %}
+      {%- endif %}
+    {% endif %}
+  {% endfor %}
+{% endif %}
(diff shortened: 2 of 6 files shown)


dbt/include/duckdb/macros/utils/external_location.sql
```diff
@@ -1,3 +1,3 @@
-{%- macro external_location(format) -%}
+{%- macro external_location(this, format) -%}
```
jwills, line 1: Does this arg need to be `this` vs., say, `relation`? Like is there any downside to renaming it, aside from the extra line in the file that needs to change?

dbt/include/duckdb/macros/utils/upstream.sql
```diff
@@ -0,0 +1,25 @@
+{%- macro register_external_upstream() -%}
+{% if execute %}
+  {% for node in model['depends_on']['nodes'] %}
+    {% set upstream = graph.nodes.get(node) %}
+    {% if upstream %}
+      {%- set upstream_rel = api.Relation.create(
+        database=upstream['database'],
+        schema=upstream['schema'],
+        identifier=upstream['alias']
```
jwills, line 9: `alias` here is intentional? (I'll admit the distinction between `identifier` and `alias` in dbt isn't always clear to me)

jwills (commented):
Looks reasonable to me, I don't think many folks have had the chance (or the desire?) to override the `external_location` macro; if we want to be extra careful we can hold the breaking change for the 1.4.0 release (which should be coming out in the next few weeks as I understand it.)

dbt/include/duckdb/macros/utils/external_location.sql
AlexanderVR: I originally had `relation` instead of `this`, but reasoned that if someone had a external_location macro and only changed the signature to match instead of changing `this` to `relation` within the macro definition, then very confusing things might result (i.e. we would be registering in the current node as the upstream node). Hence the change to shadow `this` to avoid that possibility.

If this reasoning seems baroque, happy to change back to `relation`.

dbt/include/duckdb/macros/utils/upstream.sql
AlexanderVR: In my understanding, the `identifier` is the property of the dbt Relation (encapsulating database details), where `alias` is a property of the dbt node. The `alias` (not the node `name`) is mapped to `identifier` when a node is parsed into a Relation object:
https://github.com/dbt-labs/dbt-core/blob/eb200b46877c31469f410b7d47dfd245473ed7cb/core/dbt/adapters/base/relation.py#L226-L245

dbt/include/duckdb/macros/utils/external_location.sql
jwills: It seems baroque, yeah-- I'm not worried about someone who has the sophistication to override the macro not having the sophistication to update the reference

jwills:
@AlexanderVR sorry for the radio silence, the rona has come for me at last and took me out of commission for a couple of days. Still coughing up a storm, but happy to merge this in once that ^ comment is addressed!

dbt/include/duckdb/macros/utils/external_location.sql
AlexanderVR: sounds good, I've reset the branch to the previous commit, removing this shadowing

AlexanderVR:
@jwills I'm really sorry to hear that. Hope you feel better soon.

PR #82, opened 2023-01-25
Match refactored class names from DBT 1.4.0

droher:
The new DBT version changed a couple of internal classes and exceptions. After changing them, everything works as expected -- tests are passing and my project builds just as before.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.10'
+          python-version: '3.8'
 
       - name: Install python dependencies
         run: |
@@ -169,7 +169,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.10'
+          python-version: '3.11'
 
       - name: Install python dependencies
         run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -73,7 +73,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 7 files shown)


jwills:
@droher amazing, thank you for this! I was planning on taking the opportunity that 1.4.0 provides to bump the minimum version of DuckDB that dbt-duckdb requires and also cleanup how I handle exceptions that DuckDB throws since my current impl of that is very glib /cc @mause

droher:
Nice! Also can validate that things are running smoothly on Python 3.11 now.

jwills:
Yeah I think we should adjust the tox.ini to do python 3.8 -> 3.11 instead of 3.7 -> 3.10, right?

droher:
Agree on 3.11, but is there a reason on the duckdb side to remove 3.7? DBT's tox still has it supported.

jwills:
oh does it? My bad I thought it had been deprecated on the dbt side

jwills:
I guess EOL for it isn't for another 5 months, nbd

jwills:
@droher so that they run as part of Cl I need you to update the version lists inside of https://github.com/jwills/dbt-duckdb/blob/master/.github/workflows/main.yml as well

jwills:
(I can always cut a 1.3.5 one on a branch if that proves to be a bad idea)

AlexanderVR:
@jwills sounds fine to me.

I have actually recently hit an issue with the upstream view change when there are multiple models referencing the same external table, and threads > 1. It will cause a catalog write-write conflict.

I have a fix in the works -- basically we can to do the upstream registration using the `on-run-start` hook instead of dynamically per model.

jwills:
@droher do you have any more time to iterate on this, or would you like me to take it over?

droher:
@jwills Sorry, just got back from traveling. Just got these passing on my fork's Actions.

Issue #84, opened 2023-01-26
Support aws webidentity token credentials when interacting with s3 on duckdb 


nclaeys:
When looking through the code, I noticed that duckdb itself does not support the default aws credentials provider chain. Probably because they do not want to have external dependencies.

I am using dbt-duckdb and for the glue integration you are correctly using the credential provider chain if you do not pass the credentials explicitly. I was wondering if you also want to provide this for the s3 integration?

Next to providing the static credentials through settings properties (as exists at the moment), you could use the boto3 client to get and update credentials when they are expired. These credentials can then be used in the `DuckDBConnectionWrapper` such that they are always valid and any authentication mechanism can be used.

Example usecase for us:
we use aws irsa for our pods on kubernetes and get the [following credentials](https://docs.aws.amazon.com/eks/latest/userguide/pod-configuration.html). Basically a roleArn and a webIdentityToken file, which gets updated every hour. Before I can run dbt commands, I now need to convert these aws credentials to the correct format as is accepted by duckdb as follows: 
```
import boto3

session = boto3.Session()
AWS_ACCESS_KEY_ID = session.get_credentials().access_key
AWS_SECRET_ACCESS_KEY = session.get_credentials().secret_key
AWS_SESSION_TOKEN = session.get_credentials().token
```

Only then I can execute dbt compile ...
Apart from annoying entrypoint, this has as disadvantage that if my job would take longer than an hour, the credentials would expire...

I think it would be great it the dbt-duckdb project would do this for me. Since you already depend on boto3 the impact is minimal and in line with what you already implemented for the glue integration. What do you think?

jwills:
@nclaeys ah, interesting! Can you talk me through how this would work? I'm imagining there would need to be a separate set of credential settings provided in the DuckDBCredentials object to signal that we need to use the AWS webidentity credentials, and when we create a connection, we would need to run some code to generate the credentials that DuckDB expects to use to pass to S3 (similar to what you're doing above?)

nclaeys:
I think that the best approach is to not depend to much on the type of credentials users pass, maybe use a property use_credential_provider_chain=true or somethng. In this case just use boto3 to create a session: boto3.Session(). Since that uses the credentials provider chain, you will immediately support all ways of authenticating and getting a session. From that session, you can then use similar code as I wrote before to get the ACCESS_KEY_ID, SECRET_ACCESS_KEY and SESSION_TOKEN and add it to the cursor before executing a query.

[...]

PR #87, opened 2023-02-01
Feature/84 support aws credential provider chain

nclaeys:
In support of: #84


jwills:
@nclaeys thank you so much for this! I'm going to refactor the integration of the s3 credentials provider w/the dbt credentials class a bit tomorrow and I'll send it back your way for another look!

jwills:
@nclaeys did my take on this feature over here: https://github.com/jwills/dbt-duckdb/pull/89

it's simpler, but possibly too simple? LMK what you think!

nclaeys:
> @nclaeys did my take on this feature over here: #89
> 
> it's simpler, but possibly too simple? LMK what you think!

It looks good to me, I would just change one thing for which I added a comment.
The get_caller_identity() method is nice to check whether the credentials are invalid or expired (it will raise an error if something is wrong).
This way you get immediate and clear feedback if something is wrong, which will help users diagnose what is going on. Otherwise you often only see 403 when trying to access data.

PR #89, opened 2023-02-04
Add in support for getting AWS creds using the AWS credentials chain

### LINKED ISSUE #84: Support aws webidentity token credentials when interacting with s3 on duckdb

jwills:
Fixes #84

### DIFF
--- a/dbt/adapters/duckdb/connections.py
+++ b/dbt/adapters/duckdb/connections.py
@@ -1,7 +1,9 @@
 import atexit
 import threading
+import time
 from contextlib import contextmanager
 from dataclasses import dataclass
+from functools import lru_cache
 from typing import Any
 from typing import Dict
 from typing import Optional
@@ -53,7 +55,7 @@ class DuckDBCredentials(Credentials):
         settings = self.settings or {}
         if self.use_credential_provider:
             if self.use_credential_provider == "aws":
-                settings.update(_load_aws_credentials())
+                settings.update(_load_aws_credentials(ttl=_get_ttl_hash()))
             else:
                 raise ValueError(
                     "Unsupported value for use_credential_provider: "
@@ -62,7 +64,14 @@ class DuckDBCredentials(Credentials):
         return settings
 
 
-def _load_aws_credentials() -> Dict[str, Any]:
+def _get_ttl_hash(seconds=300):
+    """Return the same value withing `seconds` time period"""
+    return round(time.time() / seconds)
+
+
+@lru_cache()
+def _load_aws_credentials(ttl=None) -> Dict[str, Any]:
+    del ttl  # make mypy happy
     import boto3.session
 
     session = boto3.session.Session()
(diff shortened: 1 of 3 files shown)


dbt/adapters/duckdb/connections.py
```diff
@@ -38,13 +38,42 @@ class DuckDBCredentials(Credentials):
...
+                    + self.use_credential_provider
+                )
+        return settings
+
+
+def _load_aws_credentials() -> Dict[str, Any]:
+    import boto3.session
+
+    session = boto3.session.Session()
+    aws_creds = session.get_credentials().get_frozen_credentials()
```
nclaeys, line 69: I would check whether the session is valid by calling get_caller_identity(). This way users get a clear error message when their session is expired/invalid credentials instead of just unauthorized when they try to access data while running their models.

jwills:
@nclaeys thanks man, so added. Next question: is it worthwhile/safe to cache the credentials? Is the overhead of fetching them such that it's not something we would want to do on every model run?

nclaeys:
> @nclaeys thanks man, so added. Next question: is it worthwhile/safe to cache the credentials? Is the overhead of fetching them such that it's not something we would want to do on every model run?

They will expire after x time, typically an hour when running on AWS services otherwise it depends on the configuration. So you could cache them and only execute get-caller-identity for every model run to check if they are still valid. I think you should be good with only catching the ExpiredTokenException and refetch the credentials. This way you at least eliminate 1 call for every model...

jwills:
@nclaeys please take another look?

nclaeys:
> @nclaeys please take another look? I went with a five-minute cache interval for the creds, happy to adjust it per your recommendations.

This looks like a great addition. Max run of 55min for a model looks totally acceptable for me.

PR #92, opened 2023-02-10
Load upstream external models with a macro

AlexanderVR:
https://github.com/jwills/dbt-duckdb/pull/80 has the unfortunate side-effect of causing a write-write catalog conflict whenever two models have the same upstream external model and threads > 1.

We solve this by providing a macro to be used as an `on-run-start` hook, that registers the needed upstream external models based on the `selected_resources`.

Unfortunately, adapters appear unable to automatically add any on-run-start hooks, so we provide information about when to enable it in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,16 @@ default:
   target: dev
 ```
 
+### Re-running external models with an in-memory version of dbt-duckdb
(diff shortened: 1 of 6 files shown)


jwills:
Hey @AlexanderVR I'd like to take another crack at this before we give up on it and fall back to documentation; I'm thinking that we could move some (all?) of the logic from the `register_external_upstream` into a function on the adapter that was guarded by a `threading.Lock` per-relation to prevent multiple relations from trying to re-create the same upstream dep at the same time, but LMK if you already tried this and/or dismissed it for some reason.

AlexanderVR:
I have not tried that but it sounds possible. Happy to leave this for now
if you want to try it out!

Didn’t originally explore that route because the on run start seemed easier
- but that was before we realized there is no good way to hook it in by
default.

PR #96, opened 2023-02-11
Allow flexible configurability of extensions

danthelion:
I was playing around with the SQLite extension which requires to an extra function call to attach the SQLite database so the tables are available in DuckDB as views. This[ function call ](https://duckdb.org/docs/extensions/sqlite_scanner#usage) requires the path of the SQLite database as a parameter which in the current version cannot be added to the profile. 

This prompted a little refactoring to allow flexible definitions & configuration of extensions.

Example profile with extensions that allow configuration:

```yaml
default:
  outputs:
   dev:
     type: duckdb
     path: /tmp/dbt.duckdb
     extensions:
      - parquet
      - sqlite:
          path: /path/to/sqlite.db
      - postgres:
          db_name: postgres
          host: localhost
          user: postgres
          source_schema: public
          sink_schema: abc
  target: dev
```

Factoring out the extensions into a separate class would allow validations to be created and other extension-specific functionality.

I'm just getting started with DuckDB and dbt adapters in general so would love some guidance on how this could be better handled! If we align on a direction I'll add tests & docs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# IntelliJ project files
+.idea
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/dbt/adapters/duckdb/connections.py
+++ b/dbt/adapters/duckdb/connections.py
@@ -8,11 +8,13 @@ from typing import Any
 from typing import Dict
 from typing import Optional
 from typing import Tuple
+from typing import Union
 
 import duckdb
 
 import dbt.exceptions
 from dbt.adapters.base import Credentials
+from dbt.adapters.duckdb.extension import ExtensionFactory
 from dbt.adapters.sql import SQLConnectionManager
 from dbt.contracts.connection import AdapterRequiredConfig
 from dbt.contracts.connection import AdapterResponse
@@ -28,7 +30,7 @@ class DuckDBCredentials(Credentials):
     path: str = ":memory:"
 
     # any extensions we want to install and load (httpfs, parquet, etc.)
-    extensions: Optional[Tuple[str, ...]] = None
+    extensions: Optional[Tuple[Union[str, Dict[str, Dict[str, str]]], ...]] = None
 
     # any additional pragmas we want to configure on our DuckDB connections;
     # a list of the built-in pragmas can be found here:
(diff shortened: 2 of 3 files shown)


jwills:
@danthelion super interesting, and really nice code! TBH I was thinking that what we need is a generic set of `attach`-related settings in the config (including the new `ATTACH` function for additional DuckDB files themselves!) 

https://duckdb.org/docs/sql/statements/attach.html

Like for this stuff I would rather have a list of `attach`-related settings that would include the type of the attachment (`duckdb`, `sqlite`, `postgres`, etc., etc.) and the config settings that each type needed so that you could attach as many different files as you needed for the run.

danthelion:
I see, so should all of the extension-level configuration go to `settings` or is there any sense in the proposed extension implementation? Not exactly sure about how the extension ecosystem works for DuckDB or what the plans are for its future.

As for the `attach`-related settings, if I understand you correctly you mean something like this:

```yaml
default:
  outputs:
   dev:
     type: duckdb
     path: /tmp/dbt.duckdb
     extensions:
      - parquet
      - sqlite
      - postgres
     settings:
      attach:
      - type: duckdb
        params:
          path: /path/to/duck.db
          alias: attached_duckdb
      - type: sqlite
        params:
          path: /path/to/sqlite.db
      - type: postgres
        params:
          db_name: postgres
          host: localhost
          user: postgres
          source_schema: public
          sink_schema: main
```

And running the specific `attach` command for each entry after the extension is loaded for a cursor, right?

jwills:
yeah the convention right now is that all of the settings go under `settings` as opposed to being coupled to the `extension`; since extensions can change their settings from version to version (and even which extensions ship as part of the core `duckdb` package vs. requiring to be loaded externally), I didn't feel like I wanted to have to keep up with the details of each extension's settings in here.

`ATTACH` functionality, on the other hand, is much more scoped and feels like something we can get our arms around. (And yes, thinking along the lines of what you outlined above, though I was thinking of it as a top-level property of the config instead of a sub-property of `settings`, but I can't tell if that was intentional or just a typo.)

jwills:
Should also note that I think the new `attach` stuff is possibly going to break some of dbt-duckdb: see https://getdbt.slack.com/archives/C039D1J1LA2/p1676059044358209

danthelion:
Alright thanks for the context, I'll close this PR 👍

PR #99, opened 2023-02-13
Add a lock to prevent upstream materializations in an in-memory model from stepping on one another

jwills:
@AlexanderVR can you take a look at this one when you get a chance and see if it fixes the issue with the upstream materializations stepping on one another?


jwills:
@AlexanderVR okay, this doesn't work either, but the way that it fails is useful/instructive; I _think_ what we can do is create the upstream views we need for the length of time for the model to be created, and then drop them again before the transaction commits so that there is no conflict across the different threads. Going to try it in another PR

jwills:
Okay, I am admitting defeat for now, gotta get the rest of the stuff out the door for 0.7.0. Gonna just go with the doc fix PR-- thanks @AlexanderVR !

AlexanderVR:
@jwills Apologies I could not take a look earlier. If you don't mind, can you share what was the instructive way this attempt failed? Is this because concurrent catalog writes -- whether rolled back or not -- from 2 threads for the same `create view` cause a conflict?

One other thought: I think this approach and the on-run-start hook approach both have a flaw where we are not adding these upstream relations to the dbt relation cache after creation! Just need to call `cache_added` like in the run task: https://github.com/dbt-labs/dbt-core/blob/v1.4.1/core/dbt/task/run.py#L284
Taking this a step further, If we place a lock around:
- checking the cache
- if not there: call create view and then add to cache

then I think we should have a guarantee that the create view command is executed exactly once, removing the conflict.

jwills:
Yep, it was the concurrent catalog writes causing a conflict-- and yes, I think what you're suggesting would work! I also toyed with the idea of _not_ creating the view at all and just rendering the upstream deps as file paths (similar to what we do with `external_location` for sources), but I just opted to go with your macro solution since I was in a hurry to get 1.4.0 out the door since 0.7.0 was broken for everyone.

Do you want to give the lock + check cache + create view and call `cache_added` trick a shot in a PR? I would be down to merge it if it worked!

AlexanderVR:
I don't think I'll have time this week -- this kind of change doesn't seem entirely trivial. But it's certainly something that should be cleaned up. Are you ok with the macro solution sitting around for a couple more weeks?

PR #100, opened 2023-02-13
WIP: Add ATTACH support for dbt-duckdb and DuckDB 0.7.0

jwills:
DuckDB 0.7.0 ships with support for multiple databases on the same connection, and I'd like to add support for this in two ways:

1) The `profile` should support the `ATTACH` options for an arbitrary number of database files (either DuckDB or SQLite) at the start of a dbt run,
2) It should be possible to do writes to _any_ of the attached databases (that aren't in read-only mode), which means that our DDL and DML statements need to be fully-specified as `database.schema.relation` by default. This will only work correctly for DuckDB 0.7.0 though, so for now I'm controlling this setting via a method on the `adapter` that checks the version of DuckDB we are running and does the "right" thing.

As the 0.7.x releases of DuckDB mature I will look at making including the database in the DML/DDL the default and I'll get rid of this hack I'm doing while we're in an in-between state.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,15 +34,18 @@ default:
 ````
 
 The `path` field should normally be the path to a local DuckDB file on your filesystem, but it can also be set equal to `:memory:` if you
-would like to run an in-memory only version of dbt-duckdb. Keep in mind that any models that you want to keep from the dbt run will
-need to be persisted using one of the external materialization strategies described below.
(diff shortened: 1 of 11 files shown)


jwills:
@tomsej hey man, hope you're doing well! If you have some time to take a look at this and give me your thoughts on the approach, I would be extremely grateful!

Issue #104, opened 2023-02-17
External materialization of csv requires a specific `format=csv` config, whereas parquet is inferred


gwenwindflower:
We should make these both inferred based on the extension for a more intuitive DX.

Current code
```
{{ 
    config(
        materialized='external', 
        location='reports/models/customers.csv',
        format='csv'
    )
    
}}
```

Would ideally be
```
{{
    config(
        materialized='external', 
        location='reports/models/customers.csv',
    )
}}
```

jwills:
Fixed in #112, not sure how I didn't close this when I merged it, my bad

PR #106, opened 2023-02-22
feat: inherit defaults for file formats from location

jpmmcneill:
solves: https://github.com/jwills/dbt-duckdb/issues/104

Also, I think there is some better handling around location & format (I think one should be required but not both). Happy to submit some changes in a PR.

Originally I was planning on adding json support to the external materialization @jwills, but based on the code I think that's already supported? Happy to add changes to the readme if this is the case.

### DIFF
--- a/dbt/include/duckdb/macros/materializations/external.sql
+++ b/dbt/include/duckdb/macros/materializations/external.sql
@@ -6,6 +6,9 @@
   {%- set glue_register = config.get('glue_register', default=false) -%}
   {%- set glue_database = render(config.get('glue_database', default='default')) -%}
 
+  -- inherit format from defined location when it is not passed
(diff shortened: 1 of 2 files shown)


jwills:
@jpmmcneill thanks for this man! Apologies for the lag here, I've been on vacation and wasn't logged into GH on my phone.

I started working on a whole reworking of the `external` stuff to address #104 and also support the `partition_by` and other more exotic write options that are available in DuckDB's `COPY...` operations, but I didn't have time to finish it before vacation started: https://github.com/jwills/dbt-duckdb/compare/master...jwills_revamp_external_configs

Would love if you took a look and let me know what you think, I'm also happy to hop on a call with you when I'm back around next week? Can DM you my info in the dbt Slack.

Issue #94, opened 2023-02-10
External sources: more options & docs


sdebruyn:
I am interested in using external sources. What I am missing in either features or the documentation (if the feature already exists):
* How do I set the delimiter for my CSV file?
* How do I specify a local path?

jwills:
Ah so the request is to update https://github.com/jwills/dbt-duckdb#external-materializations-and-sources to include info on how to setup CSV external sources?

sdebruyn:
Correct :)

PR #107, opened 2023-02-23
Add further explanation to external reading and writing to README

felippecaso:
Based on [this conversation](https://getdbt.slack.com/archives/C039D1J1LA2/p1677166441609539) in dbt Slack, add detailed explanation to `external_location` being able to be provided a explicit read function.

Resolves https://github.com/jwills/dbt-duckdb/issues/94 (hopefully)


jwills:
Hey @felippecaso, thanks so much for this! I thought this block should replace the one before it (that includes the simpler `read_parquet(...)` example), since it's substantially similar but more clearly demonstrates what you're allowed to do w/the `external_location` option on sources?

felippecaso:
> Hey @felippecaso, thanks so much for this! I thought this block should replace the one before it (that includes the simpler `read_parquet(...)` example), since it's substantially similar but more clearly demonstrates what you're allowed to do w/the `external_location` option on sources?

That makes sense! I've updated the text and added some other suggestions in text structure to address some of the points mentioned in https://github.com/jwills/dbt-duckdb/issues/94

felippecaso:
@jwills after reviewing the proposed changes in resolving conflicts, I feel the contributions are already covered by recent updates to `README`, so I'm closing this PR.

For the same reason, I also believe https://github.com/jwills/dbt-duckdb/issues/94 is already solved as well.

PR #111, opened 2023-02-27
Override the ADDITIONAL_PROPERTIES setting on the DuckDBCredentials

jwills:
This will prevent typos in the `Credentials` from causing silent failures at runtime (e.g., because extensions were not actually loaded.)

Gonna solicit some feedback on merging this as there may well be cases I haven't thought of where we _would_ want arbitrary properties to be added to the profile, but I'm not sure of what they would be just yet.


jwills:
Closing this out for now

PR #113, opened 2023-03-06
fix: clear connections before tests

AlexanderVR:
There are two tests which expect no connections to be opened. This is the case when run by themselves, but running other tests in the same pytest session can affect the class-level variable `DuckDBConnectionManager.CONN`

Until this class-level variable can be fixed, we reset the connections before these affected tests.

@jwills would you prefer that before *every* test runs the connections are reset? Can instead have a pytest fixture with `autouse=True` if you would prefer a less ad-hoc, if more invasive, approach.


jwills (approved):
Ack— thank you for fixing this @AlexanderVR!

Issue #124, opened 2023-03-24
Bug in register_upstream_external_models


vlieven:
I have an issue when running a `dbt run --select` statement where external tables are referenced. An invalid upstream location is generated, resulting in the error e.g. `IO Error: No files found that match the pattern "external/root/.parquet"` when running for the config `external_root: "external/root"`.

I've identified the problem in `macros/utils/upstream.sql`, and I'll put up a PR shortly. I'm not sure how to write a test that reproduces this issue though, so I'd appreciate it if you could maybe give me some pointers there. Thanks!

Mickychen00:
Hello. I have set the following in my dbt_project.yml:

```
on-run-start:
- "{{ register_upstream_external_models() }}"
```

However, I'm encountering a runtime error with this configuration in each model:

```
{{ config(materialized='external', location='../../data/data_check/cap_ele.parquet', format='parquet') }}
```

The error message states: 

"IO Error: No files found that match the pattern '../../data/data_check/{{ this.identifier }}.parquet'."

Howerver, if I change {{this.identifier}} to a concrete string name. It works.

What did I do wrong?


#https://github.com/duckdb/dbt-duckdb/issues/227

jwills:
My best guess is that you didn't set the `external_root` property in your profile.yml, which is the root path that dbt-duckdb uses to look for external materializations (it defaults to `.`, the local directory where dbt-duckdb is running)

https://github.com/duckdb/dbt-duckdb/blob/master/dbt/adapters/duckdb/credentials.py#L114

Try setting that field in your profile to an absolute path (possibly based on an environment variable if you're running in multiple different environments) so that the external path materialization will work correctly.

Mickychen00:
Hi. So do you mean I cannot set the external folder as a relative path over the project root?

PR #125, opened 2023-03-24
Fix bug in macros/utils/upstream.sql

vlieven:
Fixes: #124

### DIFF
--- a/dbt/include/duckdb/macros/utils/upstream.sql
+++ b/dbt/include/duckdb/macros/utils/upstream.sql
@@ -16,7 +16,7 @@
           schema=upstream['schema'],
           identifier=upstream['alias']
         ) -%}
-        {%- set location = upstream.config.get('location', external_location(upstream, upstream.config)) -%}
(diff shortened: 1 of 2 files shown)


jwills:
I see the bug and it's pretty obvious; really surprised that this integration test didn't catch it tho: https://github.com/jwills/dbt-duckdb/blob/master/tests/functional/adapter/test_rematerialize.py

jwills:
cc @AlexanderVR

AlexanderVR:
Arrrg, the `dbt` comma vs space strikes again. Basically it's creating a null selection so the test always passes. This was already a hard-fought one-line fix https://github.com/jwills/dbt-duckdb/pull/109/files but the bug found its way back.

So there are two issues exposed by fixing the test here. Thanks @vlieven for finding the model vs relation bug.

There is also a minor issue of which config to use for handling e.g. the partitioning. I've submitted https://github.com/jwills/dbt-duckdb/pull/126 to handle all of these

jwills:
@AlexanderVR I assume I can close this, as the fix was included in your PR?

PR #126, opened 2023-03-26
Upstream model registration redux

AlexanderVR:
A couple more fixes for https://github.com/jwills/dbt-duckdb/pull/125

* Fix comma vs spaces causing empty test
* Really ensure the models are getting created in this test
* Use the upstream config to generate the upstream options.


jwills:
Thank you so much @AlexanderVR !

Issue #127, opened 2023-03-29
External Location: f-string value needs more than one variable


armsepehr:
I am interested in using external_location f-string with several variable but only `name`, `indicator` and `schema` are known in the f-string.  

```
sources:
  - name: external_source
    meta:
      external_location: "s3://my-bucket/{folder2}/{name}.{suffix}"
    tables:
      - name: source1
         suffix: parquet
         folder: folder1
      - name: source2
        suffix: csv
        folder: folder2
```

Additionally, I don't want to overwrite the external_location for each table name.

armsepehr:
I think it's better to distinguish between the variable inside and outside the meta variable; however, this is a good idea to simplify the variable inside the f-string compare to directly referencing such as`source.meta["folder2]`. What you said has another advantage and the previous test already passed to reference `name`, `schema` and `identifier`. Please clarify your suggestion according to address a variable inside the f-string.

jwills:
I was thinking like this (the test illustrates how it would look): https://github.com/jwills/dbt-duckdb/pull/130

PR #128, opened 2023-03-29
Fix bug in duckdb/relation.py

armsepehr:
Fix: #127 

The README is not updated at this change, so please clarify your suggestion according this change and I can modify the documentation as well.


armsepehr:
In the other PR #130, it has a better solution and I close this one.

PR #131, opened 2023-03-30
Enable testing with file DB and fix bug in connection handling

vlieven:
I noticed that the tests are only run against a `:memory:` database. This PR modifies the tests according to the [dbt documentation](https://docs.getdbt.com/guides/dbt-ecosystem/adapter-development/4-testing-a-new-adapter#running-with-multiple-profiles) to add testing profiles, allowing to switch between an in-memory and file-backed DuckDB backend (memory is kept as the default). The test can be run using a file by running `pytest --profile file`.

This change surfaces a bug in the Environment class, where the connection is prematurely closed due to the `handles` counting. I think that the `Environment.close(cursor)` behaviour is unwanted both for memory as file databases, which would make the handle counting unneeded as well. I simplified the implementation, fixing the issue, but I'm not sure if the handle counting served another purpose?

I suppose the `tox.ini` and github workflows should be updated as well to include the second testing path, but I'll let you review these changes first. Thanks!

### DIFF
--- a/dbt/adapters/duckdb/environments.py
+++ b/dbt/adapters/duckdb/environments.py
@@ -25,11 +25,10 @@ class DuckDBCursorWrapper:
 
 class DuckDBConnectionWrapper:
     def __init__(self, env):
-        self._env = env
         self._cursor = DuckDBCursorWrapper(env.cursor())
 
     def close(self):
-        self._env.close(self._cursor)
+        self._cursor.close()
 
     def cursor(self):
         return self._cursor
@@ -49,8 +48,8 @@ class Environment:
 class LocalEnvironment(Environment):
     def __init__(self, credentials: DuckDBCredentials):
         self.creds = credentials
-        self.handles = 0
         self.conn = duckdb.connect(credentials.path, read_only=False)
+
         # install any extensions on the connection
         if credentials.extensions is not None:
             for extension in credentials.extensions:
@@ -73,7 +72,6 @@ class LocalEnvironment(Environment):
                 self.conn.execute(attachment.to_sql())
 
     def handle(self):
-        self.handles += 1
         return DuckDBConnectionWrapper(self)
 
     def cursor(self):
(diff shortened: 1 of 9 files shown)


vlieven:
Happy to help! Should I update the tox and workflow, or how do you want to proceed?

dbt/adapters/duckdb/environments.py
```diff
@@ -87,16 +85,13 @@ def cursor(self):
             cursor.execute(f"SET {key} = '{value}'")
         return cursor
 
-    def close(self, cursor):
-        cursor.close()
-        self.handles -= 1
-        if self.conn and self.handles == 0 and self.creds.path != ":memory:":
+    def close(self):
+        if self.conn:
```
jwills, line 89: So I thought i needed this handling logic for the in-memory stuff to work correctly, but maybe I'm wrong?

jwills:
Realizing it's pretty late in Belgium, gonna go the merge-and-chatgpt-it route

dbt/adapters/duckdb/environments.py
vlieven: If you don't take the `self.creds.path != ":memory:"` part into account, the handling logic produces this behaviour:
1. Create connection to duckdb -- handles = 0
2. Create cursor -- handles = 1
3. Release cursor -- handles = 0
4. Release connection because handles == 0
5. Create a new cursor, but throw an error because the duckdb connection is already closed.

By putting the extra `:memory`: check, closing the connection prematurely is prevented, but the problem will then still present itself when running against a file-backed db. Skipping the check for both memory and file seems a bit redundant, so I just removed it.

I am running the tests against dbt 1.4.5 and duckdb 0.7.1, maybe the behavior was a bit different on older versions?

jwills: yeah, I'm remembering why I did this now-- https://getdbt.slack.com/archives/C039D1J1LA2/p1665015877499679

The context there is that if you're running fal alongside dbt-duckdb, and one of the fal scripts needs to use the DuckDB connection, it can't access it b/c the dbt-duckdb process owns the write lock on the file as long as one of the connections is open. So the point of this tracking was to have the opportunity for dbt-duckdb to release the write lock completely whenever the last connection to the DB was closed so that the fal script would have the opportunity to take it over.

In the new world of dbt-fal, this may be less relevant and thus no longer worth keeping around, but I'd hate to break anybody who was running this as-is, so I'll solicit some feedback in the dbt Slack to see if taking it out is an issue for folks; there are some pretty obvious code clarity/perf wins to not having to open and close the connection over and over again.

PR #134, opened 2023-03-30
Wire up GH Actions to run file-based tests as well as in-memory tests on PRs

jwills:


jwills:
cc @vlieven as an FYI here

PR #135, opened 2023-04-01
Adds a remote environment for DuckDB databases running in a Buena Vista server

jwills:
(using the postgres protocol)

### DIFF
--- a/dbt/adapters/duckdb/environments.py
+++ b/dbt/adapters/duckdb/environments.py
@@ -133,7 +133,8 @@ class LocalEnvironment(Environment):
 
     def submit_python_job(self, handle, parsed_model: dict, compiled_code: str) -> AdapterResponse:
         con = handle.cursor()
-        ldf = lambda table_name: con.query(f"select * from {table_name}")
+        def ldf(table_name):
+            return con.query(f"select * from {table_name}")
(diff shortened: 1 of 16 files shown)


dbt/adapters/duckdb/environments.py
```diff
@@ -41,50 +45,101 @@ def handle(self):
...
         # Extensions/settings need to be configured per cursor
-        cursor = self.conn.cursor()
-        for ext in self.creds.extensions or []:
+        for ext in creds.extensions or []:
             cursor.execute(f"LOAD '{ext}'")
-        for key, value in self.creds.load_settings().items():
+        for key, value in creds.load_settings().items():
             # Okay to set these as strings because DuckDB will cast them
             # to the correct type
             cursor.execute(f"SET {key} = '{value}'")
```
Mause, line 88: Might be worth escaping quotes here?

jwills: I don't understand?

jwills: (Like, which quotes should I be escaping?)

Mause: In the case that someone passes a string containing single quotes I mean - they should be doubled up to escape them. It's not directly relevant to this pr, just figured it was worth mentioning

PR #138, opened 2023-04-04
only return non-empty creds to prevent code 400 when session token is empty

N-Clerkx:
When testing out the credential provider functionality, external models pushed to S3 failed with the following error:

```
Invalid Error: IO Error: Unable to connect to URL "https://blabla.s3.amazonaws.com/some/file/name.parquet": 400 (Bad Request)
```
This was caused by boto3 not returning a session token in my setup and dbt-duckdb setting an empty string as variable in duckdb.
Now fixed by only passing the credentials with a value.


jwills:
Thank you @N-Clerkx!

PR #140, opened 2023-04-07
Set unique_field in credentials.py

aranke:
Required in dbt 1.5+ to avoid `NotImplementedError` raised here: https://github.com/dbt-labs/dbt-core/blob/3cee9d16fa39012e30a02c66770d6b6c7a445f6c/core/dbt/contracts/connection.py#L141

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,4 @@ target/
 .python-version
 
 .DS_Store
+.idea/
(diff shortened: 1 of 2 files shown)


jwills:
@aranke thanks for this; I think the closest thing to a unique field for tracking purposes is either 1) `remote.host + remote.port` (if they're defined) or 2) `path + external_root` (if they're not.)

PR #141, opened 2023-04-08
Create a simple plugin system for loading data from external sources

### LINKED ISSUE #137: dbt-duckdb should know about Excel
Essentially, any external file format or storage system that can read/write a `pandas.DataFrame` (or an arrow record reader, or a polars data frame, or a JSON/CSV file) and that has Python library support should be usable as either a) a source or b) an external materialization target in dbt-duckdb.

Reference: [dbt-excel.com](http://dbt-excel.com)

jwills:
Fixes #137 and addresses some longstanding roadmap items (esp. to make it easier to use dbt-duckdb with data in Iceberg/delta tables.)

The idea here is that there are certain kinds of external sources (e.g., Excel files, Google Sheets, or Iceberg tables), where we need to execute a _little_ bit of custom Python code to transform the data that the source contains into something that DuckDB can consume as a table (like a pandas/polars DataFrame, or a PyArrow table/dataset.)

One way to include data from these sources in a dbt-duckdb project is to use Python models, but I've noted in my own work that this is fairly tedious to do because it involves:

1) a lot of trivial and boilerplate Python that isn't very much fun to write, and
2) some tricky and/or non-obvious code to allow the Python models to access/operate on shared resources (e.g., a Google client or an Iceberg catalog) that requires some amount of configuration info that is really best contained inside of the `profiles.yml` file that is designed for exactly that sort of thing.

The idea here is to create a plugin system for dbt-duckdb (which is, somewhat awkwardly, itself a plugin for dbt-core) which lets us define the code we need to extract data from these external systems in a way that makes them accessible to DuckDB and our downstream transformation logic. To support this, we are using some special `meta` tags on the source (ala the `external_location` trick we use now for CSV/Parquet files) as arguments to the plugin's `load` function, which then returns a data object that DuckDB knows how to convert into a table for use by the rest of the pipeline.

You can think of these plugins as a way to support "elT" use cases (as opposed to dbt's standard "ELT" workloads, where you really do want a high-powered extracting/loading system at your disposal), where the extraction/loading work is simple and safe enough that a small python script is all we want/need.


jwills:
@JCZuurmond curious for your take here, as my inspiration for this. ;-)

dbt/adapters/duckdb/plugins/gsheet.py
```diff
@@ -0,0 +1,54 @@
...
+            doc = self._gc.open(source_config.meta["title"])
+        elif "key" in source_config.meta:
+            doc = self._gc.open_by_key(source_config.meta["key"])
+        elif "url" in source_config.meta:
+            doc = self._gc.open_by_url(source_config.meta["url"])
+        else:
+            raise Exception("Source config did not indicate a method to open a GSheet to read")
+
+        sheet = None
+        if "worksheeet" in source_config.meta:
```
buremba, line 41: maybe triple `eee` is a typo?

buremba:
I wonder how useful it would be to combine it with [Singer taps](https://github.com/singer-io/getting-started/blob/master/docs/RUNNING_AND_DEVELOPING.md#a-python-tap) to support. It seems feasible by mocking `singer.write_schema` and `singer.write_records` but I didn't try it myself.

jwills:
> I wonder how useful it would be to combine it with [Singer taps](https://github.com/singer-io/getting-started/blob/master/docs/RUNNING_AND_DEVELOPING.md#a-python-tap) to support a number of different sources out of the box. It seems feasible by mocking `singer.write_schema` and `singer.write_records` and converting the data to DataFrame internally but I didn't try it myself.

Yeah this is a great q b/c I've been wondering about how far I can push this pattern. Right now, I think that these plugins run single-threaded as part of the graph compilation piece of the dbt execution, so I didn't want to make it too easy to inject things that are arbitrarily slow/complicated in here (e.g., pulling in a lot of data from the GH API) b/c that really should be done externally to dbt-duckdb using one of the many excellent EL frameworks out there which can e.g. run multiple threads/processes, restart from failure, be aware of things like API rate limits, etc., etc.

dbt/adapters/duckdb/plugins/gsheet.py
jwills: yay, thank you! What I get for not testing this plugin out yet ;-)

dbt/adapters/duckdb/environments.py
```diff
@@ -159,6 +179,15 @@ def ldf(table_name):
         self.run_python_job(con, ldf, parsed_model["alias"], compiled_code)
         return AdapterResponse(_message="OK")
 
+    def load_source(self, plugin_name: str, source_config: SourceConfig):
+        df = self._plugins[plugin_name].load(source_config)
+        assert df is not None
+        handle = self.handle()
+        cursor = handle.cursor()
+        cursor.execute(f"CREATE OR REPLACE TABLE {source_config.table_name()} AS SELECT * FROM df")
```
JCZuurmond, line 187: I love this DuckDB trick! Great way of generalising the approach

dbt/adapters/duckdb/plugins/excel.py
```diff
@@ -0,0 +1,18 @@
...
+from ..utils import SourceConfig
+
+
+class ExcelPlugin(Plugin):
+    def __init__(self, config: Dict):
+        self._config = config
+
+    def load(self, source_config: SourceConfig):
+        ext_location = source_config.meta["external_location"]
+        ext_location = ext_location.format(**source_config.as_dict())
```
JCZuurmond, line 16: This allows dynamically inserting values from the config using a format string. Could you give an example of how this is intended to be used? Does this, for example, contain the seed directory?

dbt/adapters/duckdb/plugins/excel.py
```diff
@@ -0,0 +1,18 @@
...
+        source_location = pathlib.Path(ext_location.strip("'"))
+        return pd.read_excel(source_location)
```
JCZuurmond, line 18: Let's add fetching the `sheet_name` from the config

JCZuurmond (approved):
@jwills : This is some great stuff! Very flexible and powerful.

dbt/adapters/duckdb/buenavista.py
```diff
@@ -42,5 +46,16 @@ def submit_python_job(self, handle, parsed_model: dict, compiled_code: str) -> A
         handle.cursor().execute(json.dumps(payload))
         return AdapterResponse(_message="OK")
 
-    def get_binding_char(self) -> str:
-        return "%s"
+    def load_source(self, plugin_name: str, source_config: utils.SourceConfig):
```
JCZuurmond, line 49: Could you explain what the `buenavista` class does?

PR #169, opened 2023-05-20
Add a fast-path option for loading dbt's seed files via a DuckDB COPY operation

### LINKED ISSUE #158: Seed materialization is significantly slower than using `select * from 'filename.csv'`.
The issue is that the default seed materialization for dbt involves loading everything into memory in the Python runtime, and does row by row inserts. DuckDB's builtin csv behavior is significantly faster than this approach.

The appropriate solution here is to override the implementation of the seed materialization. The 2 tricky parts may be (1) avoiding not just the inserts but the csv loading entirely, and (2) backwards compatibility.

jwills:
Fixes #158


jwills:
As is ever the case, I spent an enormous amount of time figuring out how to do this in a backwards-compatible way using as few lines of code as possible.

dwreeves:
👀

PR #173, opened 2023-05-31
Create a simple plugin system for writing data to external destinations

### LINKED ISSUE #143: Create a simple plugin system for writing data to external destinations

jwills:
Fixes #143

This implements a write-side plugin system that can run after a relation is materialized (either in DuckDB or externally) and can be used to take some action on the resulting data. I'm starting out by porting the Glue database stuff (which can persist an output parquet file as a relation in the AWS Glue catalog) to guide the implementation of the plugins to act as a dual to the source-side plugins.

One rub here is that I don't have a good functional test of the glue stuff, so I'm going to create one in a separate PR that runs against the existing glue impl, merge it, and then port the impl over to confirm that it will work against my new impl here (and likely fix the bugs I find in the process.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ coverage.xml
 *,cover
 .hypothesis/
 test.env
+.mypy_cache/
 
 # Translations
 *.mo
--- a/dbt/adapters/duckdb/environments/__init__.py
+++ b/dbt/adapters/duckdb/environments/__init__.py
@@ -10,6 +10,7 @@ import duckdb
 from ..credentials import DuckDBCredentials
 from ..plugins import BasePlugin
 from ..utils import SourceConfig
+from ..utils import TargetConfig
 from dbt.contracts.connection import AdapterResponse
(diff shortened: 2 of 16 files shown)


jwills:
Going to merge this and then start in on revamping the source-side plugins to move them into `dbt seed`

JCZuurmond:
Awsome work @jwills! Do have or know of any examples of this functionality?

JCZuurmond:
sqlachemy would be nice! 

I want to start with a reference use case, maybe as a blog post or talk, to show the Excel capabilities and to inspire other plugin ideas.

PR #187, opened 2023-06-19
Add an example of doing a write-side operation in the sqlalchemy plugin

jwills:
Fixes @JCZuurmond comment on #173


JCZuurmond:
Awesome stuff again, @jwills! 🎉

PR #214, opened 2023-07-12
fix: remove parenthesis from md_connect call

gmontanola:
image

The syntax when `token` is specified is working as intended.


jwills:
Ack-- thank you @gmontanola!

PR #228, opened 2023-08-08
Make source tags available to plugins through the SourceConfig

AlexanderVR:
I have a custom plugin where I'd like to use the source's tags to inform the plugin `load` method. This MR adds the source tags as an additional SourceConfig attribute.

Additionally, when attempting to ensure existing plugin tests passed, ran into the issue that tests that should be skipped when the test profile is not `md` would still load the motherduck plugin and attempt a connection. This is happening because any test classes using the dbt `project` fixture will execute it before the `skip_by_profile_type` fixture, as the former is at the higher class-level scope, thus rendering the `pytest.mark.skip_profile` decorator ineffectual

Fix: changed `skip_by_profile_type` autouse fixture from function to class scope.

### DIFF
--- a/dbt/adapters/duckdb/utils.py
+++ b/dbt/adapters/duckdb/utils.py
@@ -1,6 +1,7 @@
 from dataclasses import dataclass
 from typing import Any
 from typing import Dict
+from typing import List
 from typing import Optional
 from typing import Sequence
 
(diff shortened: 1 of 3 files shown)


jwills:
@AlexanderVR `tags` stuff LGTM obvi, but I'm puzzled by this MD test failure, does it make sense to you?

jwills:
Ack okay looking more closely it looks like a token provided error, which shouldn't actually be an issue, I bet it's some dumb GH thing I haven't configured correctly

PR #230, opened 2023-08-08
Add s3 support to Excel plugin through env variables

felippecaso:
Ref [this](https://getdbt.slack.com/archives/C039D1J1LA2/p1691409607134479) Slack thread, pull `settings` from profile to be available for Excel plugin (they are currently only registered directly inside `duckdb`).

### DIFF
--- a/dbt/adapters/duckdb/plugins/excel.py
+++ b/dbt/adapters/duckdb/plugins/excel.py
@@ -1,8 +1,7 @@
 import os
+import pathlib
 from typing import Any
 from typing import Dict
-
-import pathlib
 
 import pandas as pd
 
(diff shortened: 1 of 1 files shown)


dbt/adapters/duckdb/plugins/excel.py
```diff
@@ -7,9 +11,22 @@
 
 
 class Plugin(BasePlugin):
+    def initialize(self, plugin_config: Dict[str, Any]):
+        # Pass s3 settings to plugin environment
+        if 's3_access_key_id' in plugin_config:
```
jwills, line 16: linting error that these and other string literals should be quoted with `"` instead of `'`

felippecaso: My bad! Replaced them

jwills:
Thank you @felippecaso!

Issue #213, opened 2023-07-10
Allow Google Sheets plug-in to select data by range as well


Thootje:
Hi! First off, thanks for the great work on this adapter!

I'm using dbt-duckdb to fetch data from Google Sheets and then transform and output it. Apart from being able to select an entire sheet, which is done using the `get_all_records` [method](https://github.com/jwills/dbt-duckdb/blob/1f92552271bdee06df18db00cbbbf5641befe72a/dbt/adapters/duckdb/plugins/gsheet.py#L55), I'd also like to select a range instead of all records. 

gspread offers this feature through the `.get()` [method](https://docs.gspread.org/en/v5.10.0/user-guide.html#getting-a-cell-value). You can pass a range using A1 notation inside that method to select a certain range instead of the entire sheet.

I'd be happy to make a PR for this if needed.

jwills:
A PR would be most welcome— thank you!

PR #233, opened 2023-08-11
Add option to fetch Gsheet data using range

Thootje:
Related to: #213 

Finally came around to make this PR to add the option to fetch Google Sheet data using a range instead of fetching the entire sheet. 

It works by calling `.get(range)` on the sheet instead of `.get_all_records()` if a range is provided.  Within the source .yml file, you have to add a range using the `range` key and, optional, headers under the `headers` key you want to use in case your selected range doesn't have any headers. 

```yml
version: 2

sources:
  - name: test_table
    schema: output
    meta:
      plugin: gsheet
      key: some_id_here
      worksheet: data
      range: A1:B2
      headers:
        - column_a
        - column_b
    tables:
      - name: test_table
```

Finally if you don't have a matching amount of headers and number of columns in the selected range, it will give a nice exception. For example, the below source .yml fetches 3 columns, but only 2 headers are provided. This will be caught by the exception.

```yml
version: 2

sources:
  - name: test_table
    schema: output
    meta:
      plugin: gsheet
      key: some_id_here
      worksheet: data
      range: A1:C2
      headers:
        - column_a
        - column_b
    tables:
      - name: test_table
```

### DIFF
--- a/dbt/adapters/duckdb/plugins/gsheet.py
+++ b/dbt/adapters/duckdb/plugins/gsheet.py
@@ -52,4 +52,21 @@ class Plugin(BasePlugin):
         else:
             sheet = doc.sheet1
 
-        return pd.DataFrame(sheet.get_all_records())
+        if "range" in source_config:
+            range = source_config["range"]
+            df = pd.DataFrame(sheet.get(range))
+            if "headers" in source_config:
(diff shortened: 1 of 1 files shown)


jwills:
Thank you @Thootje!

PR #254, opened 2023-09-19
Refactor render_write_options for direct option rendering

thfmn:
Removed an unnecessary if-else condition in the render_write_options macro, eliminating the need for type checking.


jwills:
Thank you @thfmn!

PR #260, opened 2023-09-28
Update Excel plugin with output support

JCotton1123:
Example profiles.yaml:

```yaml
     plugins:
        - module: excel
          alias: accounts_file
          config:
            output:
              engine: 'xlsxwriter'
              engine_kwargs:
                # See https://xlsxwriter.readthedocs.io/workbook.html#Workbook
                options:
                  # See https://xlsxwriter.readthedocs.io/working_with_memory.html#memory-perf
                  strings_to_numbers: true
                  remove_timezone: true
                  use_zip64: true
                  nan_inf_to_errors: true
              file: "files/accounts.xlsx"
              index: false  # Don't output row number
              header_styling: false  # Disable default Pandas header styling
```

Example model:

```
{{ config(materialized='external', plugin='accounts_file', overrides={'sheet_name': 'New Accounts'}) }}

SELECT *
FROM {{ ref('accounts') }}
WHERE status = 'NEW'
```

### DIFF
--- a/dbt/adapters/duckdb/plugins/excel.py
+++ b/dbt/adapters/duckdb/plugins/excel.py
@@ -1,16 +1,28 @@
 import os
 import pathlib
+from threading import Lock
 from typing import Any
 from typing import Dict
 
 import pandas as pd
+from pandas.io.formats import excel
 
 from . import BasePlugin
+from . import pd_utils
 from ..utils import SourceConfig
+from ..utils import TargetConfig
+from dbt.logger import GLOBAL_LOGGER as logger
 
 
 class Plugin(BasePlugin):
     def initialize(self, plugin_config: Dict[str, Any]):
+        self._config = plugin_config
+
+        if "output" in plugin_config:
+            self._excel_writer_create_lock = Lock()
+            assert isinstance(plugin_config["output"], dict)
+            assert "file" in plugin_config["output"]
+
         # Pass s3 settings to plugin environment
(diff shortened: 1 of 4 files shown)


dbt/adapters/duckdb/plugins/excel.py
```diff
@@ -29,3 +40,45 @@ def load(self, source_config: SourceConfig):
...
+
+
+    def store(self, target_config: TargetConfig):
+        plugin_output_config = self._config["output"]
+
+        # Create the writer on the first instance of the call to store.
+        # Instead if we instantiated the writer in the constructor
+        # with mode = 'w', this would result in an existing file getting
+        # overwritten. This can happen if dbt test is executed for example.
+        if not hasattr(self, '_excel_writer'):
```
jwills, line 52: Q: does this bit need a lock? I'm assuming it's possible for multiple threads to hit it at the same time.

JCotton1123:
@jwills the one thing this PR is missing are tests. If you can provide some guidance there I am happy to implement some. Also, thanks again for taking the time to chat this morning.

JCotton1123:
@jwills just getting back to this...I attempted to follow your suggestion for implementing a test. LMK if this needs any additional changes?

JCotton1123:
@jwills I believe those issues should be resolved 🤞

jwills:
@JCotton1123! Yay-- thanks again!

Issue #241, opened 2023-09-01
Support for delta 


Yacobolo:
Looking forward to the support for delta. This would enable us the run a poor man's data lakehouse! Do you need any help? What is the eta - this year?

jwills:
Ack, sorry for the lag here @Yacobolo, I was on the road and missed this going by. I would really like to have a plugin that supported Delta akin to the one I have for [Iceberg](https://github.com/jwills/dbt-duckdb/blob/master/dbt/adapters/duckdb/plugins/iceberg.py); I'm assuming it would use the [deltalake](https://pypi.org/project/deltalake/) python package, but I personally don't have access to a Delta lake instance and tbh don't really care enough about learning how to setup a real one to do it myself "for fun."

However, if *you* (or anyone else!) does have a Delta lake instance and you know it should be configured as a dbt-duckdb plugin, I would most definitely be happy to merge it in.

milicevica23:
Hi, @jwills, I would like to try this integration. This would be my first contribution, so I would appreciate some help and guidance at the beginning. 

I did a first draft of read plugin integration [here](https://github.com/milicevica23/dbt-duckdb/tree/feature/delta-plugin)  
and doing parallel an example project [here](https://github.com/milicevica23/dbt-duckdb-delta-plugin-demo) where i showcase it 

[Here](https://github.com/milicevica23/dbt-duckdb-delta-plugin-demo/blob/main/dbt_duckdb_delta/models/source.yml) is the source configuration which loads data as the source with file and projection prunning

What workflow works best for you that you are able to give a feedback?

jwills:
Hey @milicevica23, thanks so much for taking a crack at this!

The code as-written makes sense to me, but I have to be honest that I don't have a great sense for how folks actually use the `deltalake` python module in the real world-- like, do folks really use delta tables w/o a catalog? https://delta-io.github.io/delta-rs/python/usage.html#loading-a-delta-table

milicevica23:
The nice thing is that you can but should not use a catalog to know where your table is and i thought to implement support for both ways. Or at least try to do it..
You can think of that as that we add a new file format to external files and not everybody who is on prem or doing simple projects have catalogs. But would be happy to hear feedback from others

Yacobolo:
Same here, the main use case is not the catalog, but more the metadata it generates together with the ACID transactions and time travel 🔥

jwills:
[...]

PR #263, opened 2023-10-01
feat: support reading delta tables with delta plugin

milicevica23:
add delta table plugin 
#241
an example project can be found [here](https://github.com/milicevica23/dbt-duckdb-delta-plugin-demo)


milicevica23:
Hi @jwills, i have question regarding execution lifetime

As i understand we load/create materialization in the compile time from the first node which is using source? Therefore if i register the df in that time it will/is not visible to the context of the node creation time. 

```bash
####### Here starts compilation
21:03:24  Began running node model.dbt_duckdb_delta.customer_raw
21:03:24  1 of 1 START sql table model main.customer_raw ................................. [RUN]
21:03:24  Acquiring new duckdb connection 'model.dbt_duckdb_delta.customer_raw'
21:03:24  Began compiling node model.dbt_duckdb_delta.customer_raw
##### We register df here 
SourceConfig(name='customer', identifier='customer', schema='main', database='dbt', meta={'delta_table_path': '/home/aleks/git/my-projects/banchmark-utils/utils/sf1/delta/customer', 'plugin': 'delta'}, tags=[])
21:03:24  Writing injected SQL for node "model.dbt_duckdb_delta.customer_raw"
21:03:24  Timing info for model.dbt_duckdb_delta.customer_raw (compile): 23:03:24.439860 => 23:03:24.506147
21:03:24  Began executing node model.dbt_duckdb_delta.customer_raw
21:03:24  Writing runtime sql for node "model.dbt_duckdb_delta.customer_raw"
21:03:24  Using duckdb connection "model.dbt_duckdb_delta.customer_raw"
######HERE STARTS node creation and ends compilation
21:03:24  On model.dbt_duckdb_delta.customer_raw: BEGIN
21:03:24  Opening a new connection, currently in state init
21:03:24  SQL status: OK in 0.0 seconds
21:03:24  Using duckdb connection "model.dbt_duckdb_delta.customer_raw"
21:03:24  On model.dbt_duckdb_delta.customer_raw: /* {"app": "dbt", "dbt_version": "1.6.3", "profile_name": "dbt_duckdb_delta", "target_name": "dev", "node_id": "model.dbt_duckdb_delta.customer_raw"} */
###### We df need it here 
    create  table
      "dbt"."main"."customer_raw__dbt_tmp"
  
    as (
      

SELECT * FROM "dbt"."main"."customer"
    );
```
The reasoning why i want that is because if we can somehow register that df in the execution from creation then we dont have to materialize the loading data earlier and we can leverage filter pruning from the first query automatically which will reduce drastically complexity and table preparation

Do you have some resource where i can understand execution lifecycle of the dbt <-> adapter execution and can you point me on some entry points of that interaction

jwills:
Hey @milicevica23 sorry for the lag here, I'm at a conference this week.

So at the time the source node in the dbt graph is created, we call the source plugin if one is defined (which is done here: https://github.com/duckdb/dbt-duckdb/blob/master/dbt/adapters/duckdb/relation.py#L42 ) which kicks off this logic block here: https://github.com/duckdb/dbt-duckdb/blob/master/dbt/adapters/duckdb/environments/local.py#L82

By default, that logic will materialize the data frame (or pyarrow table/dataset, or whatever) into the DuckDB database associated with the run (unless a table with the expected name already exists in the database, etc.-- there is some other logic in there for overwriting existing tables if they exist, etc., etc.)-- _but_ you can also choose to materialize the data frame as a view instead, which I think is what you're alluding to here-- i.e., we can delay actually reading the data from the df/pyarrow dataset so that you can do predicate pushdown etc. later on in the dbt run.

milicevica23:
The problem is that i tried to register df in this moment but in the very next when we materialize the node the df instance is not there. 
I have to do that registration of the dataset in the moment/session when node which is dependent of the source is created. 
I am sorry if i can't explain it right because i miss a bit of knowlage for the general adapter <-> dbt logic and when is what build. Do you have something to recommend to learn a general concept?

milicevica23:
will try to prepare a test, but have to do it tommorow

Just random two thoughts, i write it down not to forget:
1. If we assume that we do the whole process in :memory: as stated in the main readme-> meaning two calls have to be in the session? because if we create a view in one and use it in another (as currently) this view is still visible. (but scope can be not vissible but if i register it to the same session it should?) as they recommended [here ](https://stackoverflow.com/questions/77211682/dataframe-not-accessable-from-another-context-if-captured-in-a-view)
2. if we assume that two different processes are executed for creation of the source node and creation of the node which is depending on the source -> we have to move creation/registration of that df in the second process/session

milicevica23:
I added a new test which showcase how the extension is used and it throws the same error
While doing that i also tried with df registration but this doesnt work either. I need more understanding how the connection instances are build and copied around 
My feeling what happens is following:

```python
import duckdb
from deltalake import DeltaTable
import pandas as pd

conn = duckdb.connect(":memory:")
conn1 = conn.cursor()

def create_view(conn):
    df = pd.DataFrame({"x": [1, 2, 3]})
    conn.register("test_materialized",df)
    return conn

conn = create_view(conn)

conn.sql("SELECT * FROM test_materialized").show()
conn1.sql("SELECT * FROM test_materialized").show()

conn.close()
conn1.close()
```
This throws:
```
CatalogException: Catalog Error: Table with name test_materialized does not exist!
Did you mean "sqlite_master"?
```
We create an connection in one point and then duplicate it but we dont register our df in the same connection instance

You did a very nice job with intercepting the creation of the node and adding plugins i like that idea a lot

milicevica23:
Hi @jwills i made it work as i wanted
Now the following sql query 
```sql
SELECT c_custkey, c_nationkey 
FROM {{source("delta_source", "customer")}}
where c_nationkey = 15
```
Load just the needed parts into the memory and filter and selection are automatically pushed down from the query. Explane produce following:
                   
┌─────────────
│        ARROW_SCAN         
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ 
│         c_custkey         
│        c_nationkey       
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ 
│Filters: c_nationkey=15 AND
│   c_nationkey IS NOT NULL
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ 
│           EC: 1           │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ 
│            5921           │
│          (0.00s)          │
└─────────────

The problem was as i described above, between runs the registration df was not copied with the new coursor. There are multiple similar [issues](https://github.com/duckdb/duckdb/issues/1736)

I now register the df by the load time in the local environment and reregister it in each connection before execute. I will try to understand if i can move it to some other point as for example initialization of the cursor. 
I would be happy if you can look over code and provide some feedback. 

I still have to test if the remote storage as azure and s3 works out of box but it is slowly getting ready

.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
```
jwills, line 5: FYI that I'm not going to want to commit this file in the PR (though I obviously understand why you have it running locally)

dbt/adapters/duckdb/environments/local.py
```diff
@@ -15,7 +16,12 @@ def __init__(self, cursor):
     def __getattr__(self, name):
         return getattr(self._cursor, name)
 
-    def execute(self, sql, bindings=None):
+    def execute(self, sql, bindings=None):   
+        #register all dfs 
+        #TODO is this okey to be here?
```
jwills, line 21: Yep I see what's going on here; will think of a place that we can move this block so that the registered DFs are available on the cursor via a call to the plugin

jwills: So I think what we're going to need to do is to add a `configure_cursor` method to the `BasePlugin` interface and then call it in here: https://github.com/duckdb/dbt-duckdb/blob/master/dbt/adapters/duckdb/environments/__init__.py#L109

...that way the plugins can keep track of their own data frames and register them on each cursor as they get created (you can see from the `initialize_cursor` method that we have the same issue that the settings on each cursor do not persist from the parent connection to the child)

.vscode/launch.json
milicevica23: i deleted it, didn't mean to let it be there at the very end just used for transition between computers

dbt/adapters/duckdb/environments/local.py
milicevica23: So i made a bit refactoring as you said. I think it makes a lot of sense 

I had a problem that in the moment when i want to cretate my view i have to have different name of my df. So i moved creation here

dbt/adapters/duckdb/environments/local.py
```diff
@@ -100,20 +102,29 @@ def load_source(self, plugin_name: str, source_config: utils.SourceConfig):
...
                     # Nothing to do (we ignore the existing table)
                     return
         df = plugin.load(source_config)
         assert df is not None
-        materialization = source_config.meta.get("materialization", "table")
-        cursor.execute(
-            f"CREATE OR REPLACE {materialization} {source_config.table_name()} AS SELECT * FROM df"
-        )
+
+        if plugin_name == "delta": # potentially should all plugins use configure_cursor
```
jwills, line 114: this block seems redundant now, no? Aren't all cursors getting configured by plugins?

jwills (commented):
Alright this is looking really good-- there are some formatting fixes that will need to get applied, but the logic/setup looks good/right to me

dbt/adapters/duckdb/environments/local.py
milicevica23: Hm i am not sure if i understand what you think. There are multiple things why i wanted to move that code out of general flow:
* The delta plugin always want a VIEW and not table as materialization
* I cant create a view with such a general df because i have to register views under different names -> therefore i moved the whole view creation block inside

Those are the reasons why i couldn't put my code there but just to coll configure_cursor in my opinion the plugins which wants and are okey with general flow to return df and load it with that materialization should/can use else branch and others which need more configuration go into `plugin in ("delta",....)` branch and configure each cursor. This also leaves the room for backportability that i dont have to change something by other plugins till it is not needed.     

Just a note i am not sure but i have feeling that if we for example in excel plugin allow materialization to be view that it will have the same problem as i had that df is not registered so this can maybe be a potential bug

milicevica23:
Sure, just point it out where and what i have to change and i will do it. Also excuse for my English not my first language so i will be happy if you see there some potential to improve comments or documentation just point it out and i will add it. 

So what i see that is still to do:
* add some readme documentation how to use it -> where should i do it exactly?
* add one extras in the setup in order that when you install it you install deltalake too 
* there is a small bug when you don't define schema to be **main** under the source configuration you get the error that there is no schema where the source view can be created -> i have to look into it, one friend of mine found it out, maybe will have here some questions
* add few tests and make it running 

I will try to go over those stuff in the next days, please add some points if i forgot some

dbt/adapters/duckdb/environments/local.py
jwills: Ah, I think I see the confusion-- so you cannot use a plugin (like delta) as the write-path for a regular dbt python model (which, as you noted, always requires a table.) You can only use a plugin with the dbt-duckdb specific `external` materialization, which ALWAYS creates a view in DuckDB on top of the data files that it writes. So for the write-side of delta, you want to implement the `store` method on your Plugin class, which gets called here:

https://github.com/duckdb/dbt-duckdb/blob/master/dbt/adapters/duckdb/impl.py#L80

...and it should be responsible for loading the parquet files that were written into a a delta table.

milicevica23: I am not so sure that if i understood. 
There are two topics one is your question
* Aren't all cursors getting configured by plugins? -> and after rereading if i understand it correctly you want to say that this block with configure_cursor is redudant because it will be configured by new cursor anyway. -> This is true you were right and i deleted that code block

and another one is:
* In your last comment above you are speaking about writing to the delta table? -> i would like that we do first just the reading in this PR and i have to do first a concept how to write to delta tables because i think it will not be trivial and there is also some knowlage missing on my side how actually the adapter and sql execution is interchanging and compiled. Therefore i would move that in separate PR if thats okey for you.

milicevica23:
One more note i also tried this with storage options for Azure and it works very well and out of box so i don't think that there should be some problems with other providers as S3 or GCP storage. If there are then they are in the delta lake to fix but it should work.

I also tried dbt power user and it also works out of box which i think is very nice because. This means you can profile your delta tables very simple from the vs code

dbt/adapters/duckdb/environments/local.py
jwills: Ah that is totally okay with me, thanks!

dbt/adapters/duckdb/environments/local.py
milicevica23: How do you do it normally, do you resolve such conversation or? 
By us normally is done by the person who initialized it? but otherwise i can do that too

milicevica23:
Hi @jwills, thank you for your help and all feedback.
From the logical perspective i think this should work now, i would be happy if you have time to look once again over it and say what formating can i fix. 
Do you think that we can write this in the dbt slack chat and if somebody has some feedback or something similar or we can merge it and say that is experimental and the adapt it, if somebody has some feedback?

dbt/adapters/duckdb/plugins/delta.py
```diff
@@ -0,0 +1,55 @@
...
+class Plugin(BasePlugin):
+    def initialize(self, config: Dict[str, Any]):
+        self._REGISTERED_DF: dict = {}
+    
+    def configure_cursor(self, cursor):
+        for source_table_name, df in self._REGISTERED_DF.items():
+            df_name = source_table_name.replace(".", "_") + "_df"
+            cursor.register(df_name, df)
+            cursor.execute(
+                f"CREATE OR REPLACE VIEW {source_table_name} AS SELECT * FROM {df_name}"
```
AlexanderVR, line 20: curious, why the indirection instead of just referencing `df` directly in the create view statement?

AlexanderVR (commented):
A few random comments since I might want to use this plugin :-D

dbt/adapters/duckdb/plugins/delta.py
milicevica23: There are two reasons
1. Historically code was differently structured and I was fighting to define dataframe which will persist between sessions executed by the dbt and therefore had first to  register it in another scope and then to reference it by unique name
2. I was not sure how duckdb registers that dataframe when it is defined in the create statement and especially when you have more then two tables and start to execute things in parallel so i thought unique name per source dataframe makes sense

geoHeil:
One note. from my side: the internal parquet reader is << 1 second to register the tables. The plugin-based delta reader took 5 seconds for me to process.

I do not know if there is potential to improve (given the external nature of the reader) but if possible, it would be nice to improve the performance.

milicevica23:
Interesting, I would take a look into it. I have to check if this comes from environment setup or reading metadata (by big delta tables, it can be that this takes time). Most important is that we don't load data into memory before pushdown predicates from the first materialization. 
**One note:** you have to load and persist data with this delta-adapter and not make a view over it because the instance of the delta pointer df is not in the scope anymore. So if you build a view, the duckdb file has just the representation of it but will throw an error because it doesn't know anything about the delta pointer.

dbt/adapters/duckdb/environments/local.py
```diff
@@ -100,20 +106,27 @@ def load_source(self, plugin_name: str, source_config: utils.SourceConfig):
...
                     # Nothing to do (we ignore the existing table)
                     return
         df = plugin.load(source_config)
         assert df is not None
-        materialization = source_config.meta.get("materialization", "table")
-        cursor.execute(
-            f"CREATE OR REPLACE {materialization} {source_config.table_name()} AS SELECT * FROM df"
-        )
+
+        if plugin_name not in ["delta"]: # plugins which configure cursor itselfs
```
jwills, line 118: I'm still not thrilled about special-casing the `delta` plugin like this; would it be better if we simply maintained the dictionary of source-created data frames (or dataframe-like equivalents) in the `Environment` instance, instead of inside of the `Plugin` class, for all of the sources? Would that work?

jwills (commented):
Sorry it took me a bit to get back to this, had a busy couple of days

dbt/adapters/duckdb/environments/local.py
milicevica23: In my opinion, we have to distinguish between materialization. Correct me if I am wrong
If you create table as df then you have the df instance and you copy data into duckdb process, if you create a view as df then you hold just the representation and pointer to df and load the first time when you resolve this view. 
This is why other plugins work with table materialization and without each cursor configuration. 

Therefore I am not sure if the idea is good to hold every instance of df in a dictionary in the Environment. Maybe I could make it so that we return a dataframe from the load function and if the materialization is a view register it to Environment. This could work and would solve also other plugins problems with the materialization view. 

Let me try it! Just thought about it while answerinng 😁

milicevica23: Hi @jwills i did refactor now, can you take another look?
So i moved the logic into environment and abstracted it over the plugins if the plugin sets materialization to view
What i would like to do but i am not sure where exactly is to set default materialization for delta framework to view because it is per default better but it is not so much important

jwills:
just pushed some formatting fixes; going to take a shot at a) seeing if I can get the testing running for myself and then b) see if I can simplify things a bit more while keeping the tests passing

jwills:
@milicevica23 so I think I'm good to go with this as-is; anything I'm not sure about in the implementation is stuff that we can iterate on in the future.

The _one_ last thing I think you wanted was a way to change the default materialization type of a source plugin from a `table` to a `view`, right? Do you just want to add that as a method on the `BasePlugin` class that returns `"table"` by default and then the delta plugin overrides it to return `"view"`?

jwills:
Hooray! Thank you @milicevica23! 🎉

PR #269, opened 2023-10-12
Enable batch processing on Python models

droher:
Creates a separate cursor for the `dbt` object in a Python model, which prevents deadlocks when trying to read and write at the same time.

### DIFF
--- a/dbt/adapters/duckdb/environments/__init__.py
+++ b/dbt/adapters/duckdb/environments/__init__.py
@@ -148,9 +148,11 @@ class Environment(abc.ABC):
                     "Python module spec is missing loader: {}".format(identifier)
                 )
 
+            # Create a separate read cursor to enable batched reads/writes
+            cur = con.cursor()
             # Do the actual work to run the code here
             dbt = module.dbtObj(load_df_function)
-            df = module.model(dbt, con)
+            df = module.model(dbt, cur)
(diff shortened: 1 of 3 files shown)


dbt/adapters/duckdb/environments/__init__.py
```diff
@@ -148,9 +148,11 @@ def run_python_job(cls, con, load_df_function, identifier: str, compiled_code: s
                     "Python module spec is missing loader: {}".format(identifier)
                 )
 
+            # Create a separate read cursor to enable batched reads/writes
+            cur = con.cursor()
```
jwills, line 152: So one thought-- the child cursor that we create from the `con` here won't inherit any of the settings (i.e., the `SET key = value` stuff) from its parent (this is why I have that `initialize_cursor` classmethod in the base `Environment` class. I think we will need to call that method here on the `cur` object that you just created so that it has all of the settings that the client code expects.

jwills (commented):
This looks great, just the one thing we need to fix

jwills:
also don't worry about the MD test, I need to just shut that one down while they get moved over to 0.9.1

jwills:
This is so great @droher -- thank you so much for fixing it! I'm going to cut a new release later tonight to get it out there.

PR #270, opened 2023-10-13
Fix cursor isolation on Python model

droher:
The read cursor is controlled by a separate function - changing so that the new cursor is applied to `materialize` instead.


jwills:
Thanks again-- nice catch!

PR #272, opened 2023-10-15
Re-enable returning a temp table in python models

AlexanderVR:
The following sequence of MRs added cursor isolation when writing back to Duckdb:
- https://github.com/duckdb/dbt-duckdb/pull/269
- https://github.com/duckdb/dbt-duckdb/pull/270

This has the consequence that any dbt python model that returns a DuckDbPyRelation referring to a temporary table will fail, since temp tables do not cross cursor boundaries.

We add a special code path to re-enable this case.


jwills:
Ack, thank you so much @AlexanderVR ! cc'ing @droher for his awareness as well.

PR #290, opened 2023-11-09
Add support for providing additional kwargs for sqlalchemy connection

amaralvieira:
Additional keyword arguments may vary across dialects, and some connections may not work without specific keyword arguments.

Additional keyword arguments can now be added at the same level as the connection_url.

Example:
```
default:
  outputs:
    dev:
      type: duckdb
      path: /tmp/dbt.duckdb
      plugins:
        - module: gsheet
          config:
            method: oauth
        - module: sqlalchemy
          alias: myoracledb
          config:
            connection_url: "{{ env_var('DBT_ENV_SECRET_SQLALCHEMY_URI') }}"
            thick_mode: true
        - module: path.to.custom_udf_module
```


jwills:
Ah, this is great— thank you!

PR #299, opened 2023-12-01
Fix sqlite write

### LINKED ISSUE #291: Writing to an SQLite database
Hey, great project love the versatility it offers!

I am having an issue trying to write to a SQLite database, I have the following configuration, and I get this error `SQLite databases do not support creating new schemas`.

Am I doing something stupid or is this not possible to do?

Thanks!

`dbt_project.yml`
```yaml
models:
  raddd_dbt:
    +materialized: table
    render_midis:
      +materialized: incremental
      +database: audio
```

`profiles.yml`
```yaml
raddd_dbt:
  outputs:
    dev:
      type: duckdb
      path: ../data/dev.duckdb
      threads: 1
      attach:
        - path: ../data/audio.db
          type: sqlite
  target: dev
```

Nintorac:
Fixes #291


tests/functional/plugins/test_sqlite.py
```diff
@@ -0,0 +1,67 @@
...
+    @pytest.fixture(scope="class")
+    def profiles_config_update(self, dbt_profile_target, sqlite_test_db):
+        return {
+            "test": {
+                "outputs": {
+                    "dev": {
+                        "type": "duckdb",
+                        "path": dbt_profile_target.get("path", ":memory:"),
+                        "attach": [
+                           {'path': sqlite_test_db}
```
jwills, line 45: Doesn't this need to have the `sqlite` type included? I thought the default assumed a DuckDB database was being attached

dbt/include/duckdb/macros/adapters.sql
```diff
@@ -1,7 +1,21 @@
-
 {% macro duckdb__create_schema(relation) -%}
   {%- call statement('create_schema') -%}
-    create schema if not exists {{ relation.without_identifier() }}
+    {% set sql %}
+        select type from duckdb_databases()
```
jwills, line 4: Nit-ish question: when was this method added? I want to know if we need to do a bump from the minimal duckdb >= 0.7.0 dependency we include in setup.py

tests/functional/plugins/test_sqlite.py
Nintorac: They show this approach in the `sqlite_scanner` [docs](https://github.com/duckdb/sqlite_scanner#reading-data-from-sqlite)

I believe if the file has the `.db` extension it is assumed to be sqlite

dbt/include/duckdb/macros/adapters.sql
Nintorac: Looks good

image

jwills (approved):
Amazing, thank you so much @Nintorac!

PR #326, opened 2024-01-30
Add workaround for temporary tables in remote database when running incremental model

guenp:
A few months ago an issue was raised in the [MotherDuck community Slack](https://motherduckcommunity.slack.com/archives/C058PJL4R0S/p1698421879315749) that `CREATE TEMPORARY TABLE` when using [incremental models](https://docs.getdbt.com/docs/build/incremental-models) was causing slowdowns on `dbt`.

Summary of the underlying issue:
- when using a remote MotherDuck `duckdb` instance, all temporary tables are created locally
- `dbt` incremental models create a temporary table for appending tuples to an existing table
- this means that whenever an incremental model is run on a MotherDuck database (`"md:my_database"`), a new table is created locally, tuples are transported from the server into the new local table, and then sent back to the server when the final `INSERT INTO` is executed. This adds a nontrivial round trip time causing said slowdown.

This PR contains a quick suggestion for a workaround until we support temporary tables in MotherDuck. It creates a "mock" temporary table in the database and deletes it afterwards.
Some thoughts for @jwills and @nicku33:
- The mock temp table is created in the same database/schema as the target table(s). We might want to change that? DuckDB currently stores temp tables in a schema named `temp`.
- It might make sense to move the logic to the MotherDuck extension and instead drop the table server-side whenever the connection closes so we don't have to rely on `dbt` to run successfully after a mock temp table is created

### DIFF
--- a/dbt/adapters/duckdb/environments/__init__.py
+++ b/dbt/adapters/duckdb/environments/__init__.py
@@ -166,7 +166,10 @@ class Environment(abc.ABC):
         # connection that they need to do
         if plugins:
             for plugin in plugins.values():
-                plugin.configure_connection(conn)
+                # Only connect to MotherDuck if this is not already
+                # done by default by connecting to "md:<ddb name>"
+                if not creds.is_motherduck:
+                    plugin.configure_connection(conn)
 
         return conn
 
--- a/dbt/adapters/duckdb/environments/__init__.py
+++ b/dbt/adapters/duckdb/environments/__init__.py
@@ -167,7 +167,7 @@ class Environment(abc.ABC):
         if plugins:
             for plugin in plugins.values():
                 # Only connect to MotherDuck if this is not already
-                # done by default by using "md:<ddb name>" path
+                # done by default by using "md:<db name>" path
                 if not creds.is_motherduck:
                     plugin.configure_connection(conn)
 
--- a/dbt/adapters/duckdb/environments/__init__.py
+++ b/dbt/adapters/duckdb/environments/__init__.py
@@ -166,10 +166,7 @@ class Environment(abc.ABC):
         # connection that they need to do
         if plugins:
             for plugin in plugins.values():
-                # Only connect to MotherDuck if this is not already
-                # done by default by using "md:<db name>" path
-                if not creds.is_motherduck:
-                    plugin.configure_connection(conn)
+                plugin.configure_connection(conn)
 
         return conn
 
--- a/dbt/adapters/duckdb/environments/__init__.py
+++ b/dbt/adapters/duckdb/environments/__init__.py
@@ -167,7 +167,7 @@ class Environment(abc.ABC):
         if plugins:
             for plugin in plugins.values():
                 # Only connect to MotherDuck if this is not already
-                # done by default by connecting to "md:<ddb name>"
+                # done by default by using "md:<ddb name>" path
                 if not creds.is_motherduck:
                     plugin.configure_connection(conn)
 
--- a/dbt/adapters/duckdb/impl.py
+++ b/dbt/adapters/duckdb/impl.py
@@ -1,7 +1,8 @@
 import os
+from typing import Any
 from typing import List
 from typing import Optional
-from typing import Sequence, Any
+from typing import Sequence
 
 import agate
 
@@ -26,6 +27,7 @@ from dbt.exceptions import DbtRuntimeError
 
 TEMP_SCHEMA_NAME = "temp_schema_name"
 DEFAULT_TEMP_SCHEMA_NAME = "dbt_temp"
+
 
 class DuckDBAdapter(SQLAdapter):
     ConnectionManager = DuckDBConnectionManager
@@ -226,10 +228,10 @@ class DuckDBAdapter(SQLAdapter):
             return super().render_column_constraint(constraint)
 
     def pre_model_hook(self, config: RuntimeConfigObject) -> None:
-        """A hook for reading
-        """
+        """A hook for reading"""
         self._temp_schema_name = config.model.config.meta.get(
-            TEMP_SCHEMA_NAME, DEFAULT_TEMP_SCHEMA_NAME)
+            TEMP_SCHEMA_NAME, DEFAULT_TEMP_SCHEMA_NAME
+        )
         super().pre_model_hook(config)
 
     @available
@@ -239,9 +241,7 @@ class DuckDBAdapter(SQLAdapter):
         table that is dropped at the end of the incremental macro or post-model hook.
         """
         return Path(
-            schema=self._temp_schema_name,
-            database=model.database,
-            identifier=model.identifier
+            schema=self._temp_schema_name, database=model.database, identifier=model.identifier
         )
 
     def post_model_hook(self, config: RuntimeConfigObject, context: Any) -> None:
@@ -251,11 +251,11 @@ class DuckDBAdapter(SQLAdapter):
         if self.is_motherduck():
             if "incremental" == config.model.get_materialization():
                 temp_relation = self.Relation(
-                    path=self.get_temp_relation_path(config.model),
-                    type=RelationType.Table
+                    path=self.get_temp_relation_path(config.model), type=RelationType.Table
                 )
                 self.drop_relation(temp_relation)
         super().post_model_hook(config, context)
+
 
 # Change `table_a/b` to `table_aaaaa/bbbbb` to avoid duckdb binding issues when relation_a/b
 # is called "table_a" or "table_b" in some of the dbt tests
--- a/dbt/adapters/duckdb/impl.py
+++ b/dbt/adapters/duckdb/impl.py
@@ -231,7 +231,7 @@ class DuckDBAdapter(SQLAdapter):
             return super().render_column_constraint(constraint)
 
     def pre_model_hook(self, config: Any) -> None:
-        """A hook for reading"""
+        """A hook for getting the temp schema name from the model config"""
         self._temp_schema_name = config.model.config.meta.get(
             TEMP_SCHEMA_NAME, self._temp_schema_name
         )
--- a/dbt/adapters/duckdb/impl.py
+++ b/dbt/adapters/duckdb/impl.py
@@ -42,6 +42,9 @@ class DuckDBAdapter(SQLAdapter):
         ConstraintType.foreign_key: ConstraintSupport.ENFORCED,
     }
 
+    # can be overridden via the model config metadata
+    _temp_schema_name = DEFAULT_TEMP_SCHEMA_NAME
+
     @classmethod
     def date_function(cls) -> str:
         return "now()"
@@ -230,7 +233,7 @@ class DuckDBAdapter(SQLAdapter):
     def pre_model_hook(self, config: Any) -> None:
         """A hook for reading"""
         self._temp_schema_name = config.model.config.meta.get(
-            TEMP_SCHEMA_NAME, DEFAULT_TEMP_SCHEMA_NAME
+            TEMP_SCHEMA_NAME, self._temp_schema_name
         )
         super().pre_model_hook(config)
 
--- a/dbt/adapters/duckdb/impl.py
+++ b/dbt/adapters/duckdb/impl.py
@@ -1,7 +1,7 @@
 import os
 from typing import List
 from typing import Optional
-from typing import Sequence, Mapping, Any
+from typing import Sequence, Any
 
 import agate
 
@@ -24,6 +24,7 @@ from dbt.contracts.relation import RelationType
 from dbt.exceptions import DbtInternalError
 from dbt.exceptions import DbtRuntimeError
 
+TEMP_SCHEMA_NAME = "temp_schema_name"
 DEFAULT_TEMP_SCHEMA_NAME = "dbt_temp"
 
 class DuckDBAdapter(SQLAdapter):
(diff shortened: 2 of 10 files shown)


jwills:
The approach here makes sense to me-- how are we testing this out?

dbt/include/duckdb/macros/materializations/incremental.sql
```diff
@@ -1,6 +1,9 @@
 {% materialization incremental, adapter="duckdb", supported_languages=['sql', 'python'] -%}
 
   {%- set language = model['language'] -%}
+  {%- set duckdb_path = adapter.config.credentials.path -%}
+  -- only create temp tables if using local duckdb, as it is not currently supported for remote databases
+  {%- set temporary = not duckdb_path.startswith("md:") and not duckdb_path.startswith("motherduck:") -%}
```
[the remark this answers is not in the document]
guenp: Done!

guenp:
> The approach here makes sense to me-- how are we testing this out?

Well, I'm working on adding a unit test for this, based on the manual test I ran. I noticed there are test files `test_motherduck.py` and `test_incremental.py`, but they're still empty. I think it's out of scope for this PR to implement the `test_incremental.py` fully for all backends. I can add it to `test_motherduck.py` for now. Does that sound good?

Also, I saw that the `md` `tox` environment is currently not tested in CI/CD (for obvious reasons), so I'll just aim for a local pass if that works for you.

jwills:
That sounds great— and yes, I trust that you will not break your own
product here. 😉

tox.ini
```diff
@@ -44,6 +44,7 @@ description = adapter function testing using MotherDuck
 skip_install = True
 passenv = *
 commands = {envpython} -m pytest --profile=md --maxfail=2 {posargs} tests/functional/adapter
+base_python = 3.11
```
guenp, line 47: I ran into [Error in pip install duckDB #9301](https://github.com/duckdb/duckdb/issues/9301), so had to explicitly set the Python version to 3.11 here to run the tests with `tox run -e md`

dbt/include/duckdb/macros/materializations/incremental.sql
```diff
@@ -1,6 +1,8 @@
 {% materialization incremental, adapter="duckdb", supported_languages=['sql', 'python'] -%}
 
   {%- set language = model['language'] -%}
+  -- only create temp tables if using local duckdb, as it is not currently supported for remote databases
+  {%- set temporary = not adapter.config.credentials.is_motherduck -%}
```
jwills, line 5: ah yeah I'd like this to be an adapter method (like `adapter.is_motherduck()` ) vs. exposing the whole call chain in here; I like to abstract this stuff away from the jinja as much as I can

dbt/include/duckdb/macros/materializations/incremental.sql
```diff
@@ -41,11 +43,16 @@
   {% else %}
     {% if language == 'python' %}
       {% set build_python = create_table_as(False, temp_relation, compiled_code, language) %}
+      -- drop the temp relation after we insert the incremental data into the target relation
+      {% set need_drop_temp = True %}
```
jwills, line 47: Ah this looks like a bug fix-- do I have that right? If so, thank you!

tests/conftest.py
```diff
@@ -69,7 +69,7 @@ def dbt_profile_target(profile_type, bv_server_process, tmp_path_factory):
                 "Please set the MOTHERDUCK_TOKEN environment variable to run tests against MotherDuck"
             )
         profile["disable_transactions"] = True
-        profile["path"] = "md:test"
+        profile["path"] = "md:plugin_test"
```
jwills, line 72: this will break for main b/c the "md:test" stuff is the namespace I use on my own MD account for testing purposes, so let's take it out prior to merging

dbt/adapters/duckdb/environments/__init__.py
```diff
@@ -166,7 +166,10 @@ def initialize_db(
         # connection that they need to do
         if plugins:
             for plugin in plugins.values():
-                plugin.configure_connection(conn)
+                # Only connect to MotherDuck if this is not already
```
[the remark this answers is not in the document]
guenp: Ah yes - so I had to add this if-statement because when I was running my test and connecting to a MotherDuck database (e.g. `"md:plugin_test"`, `MD_CONNECT` actually failed with the message `"already connected"`. This is because upon connecting to the database the MotherDuck extension is already loaded + database is attached automatically. I'll clarify the docstring a bit.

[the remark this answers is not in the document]
guenp: But you are right that this is not the right place for that to live - I'll move it to the DuckDB plugin specifically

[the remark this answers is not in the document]
jwills: gotcha, that totally makes sense 👍

dbt/include/duckdb/macros/materializations/incremental.sql
guenp: It's not, I introduced `need_drop_temp` here! When using `CREATE TEMPORARY TABLE` the table is dropped automatically when the `duckdb` database is closed so there's no need to drop it in the macro.

guenp:
Thanks @jwills for the review! I'll address your comments shortly. I also wanted to note that I talked to @nicku33 about this:
> The mock temp table is created in the same database/schema as the target table(s). We might want to change that?

We decided to go with a separate schema `dbt_temp` for the `temp_relation`s used by incremental models. This way users won't accidentally query them in BI tools.

jwills:
> Thanks @jwills for the review! I'll address your comments shortly. I also wanted to note that I talked to @nicku33 about this:
> 
> > The mock temp table is created in the same database/schema as the target table(s). We might want to change that?
> 
> We decided to go with a separate schema `dbt_temp` for the `temp_relation`s used by incremental models. This way users won't accidentally query them in BI tools.
> 
> What are your thoughts?

I am good with that!

jwills:
@guenp apologies to pile this on, but would you mind changing this code to also use your new `is_motherduck` function? Just realized I forgot I had introduced some duplication of this already! https://github.com/duckdb/dbt-duckdb/blob/master/dbt/adapters/duckdb/environments/local.py#L53

dbt/adapters/duckdb/environments/__init__.py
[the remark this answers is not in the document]
guenp: Looks like we don't need `MD_CONNECT` after all, just loading the extension and setting the token is sufficient!

tox.ini
guenp: It looks like I need to reverse this change for the CI to pass, see https://github.com/duckdb/dbt-duckdb/actions/runs/7720091362/job/21085376439

guenp:
> @guenp apologies to pile this on, but would you mind changing this code to also use your new `is_motherduck` function? Just realized I forgot I had introduced some duplication of this already! https://github.com/duckdb/dbt-duckdb/blob/master/dbt/adapters/duckdb/environments/local.py#L53

Ah yes, I totally forgot about that, thank you for the reminder! It's done.

I've now pushed all my changes. Mainly what I've added since you last reviewed, in addition to addressing your comments:

- `DuckDBAdapter.get_temp_relation_path(model)`. This method is used in the `incremental.sql` macro and will create a new unique name for a `temp_relation` that is to be dropped at the end of the macro, with identifier `dbt_temp.<schema_name>__<target_relation_name>`.

- `DuckDBAdapter.post_model_hook(config, context)`. This method fixes a memory leak issue that arises if the incremental macro doesn't run successfully (e.g. when it runs into a `CompilationError` as it does in [this test](https://github.com/dbt-labs/dbt-core/blob/0836095a572532fc22b1c0981346eebe56f7d1fb/tests/adapter/dbt/tests/adapter/incremental/test_incremental_on_schema_change.py#L96)) and never reaches the lines where it deletes the `temp_relation`. The method makes sure to drop the `temp_relation` if it still exists after the model failed (see the `try-except` block that intercepts this [here](https://github.com/dbt-labs/dbt-core/blob/0836095a572532fc22b1c0981346eebe56f7d1fb/core/dbt/task/run.py#L298)).

The behavior now looks something like this:
(1) While running [unit test](https://github.com/duckdb/dbt-duckdb/blob/master/tests/functional/adapter/test_incremental.py):
image
(2) After unit test is finished and all test-related and "temp" tables are cleaned up:
image

@nicku33 and I decided against dropping `dbt_temp` in the end because it might interfere with concurrent processes.
We also discussed making the name `dbt_temp` configurable. I was thinking to implement this via the model `config`, but do you know if there may be a more global way to do configure this?

jwills:
The model `config` is the right place to use for setting/overriding the temp schema name, and that setting can be set globally for all models if need be by specifying it in e.g. the `dbt_project.yml` file; see https://docs.getdbt.com/reference/model-configs#model-specific-configurations

guenp:
Oops, I didn't actually run the unit test with the default profile. 😁
It looks like it failed because of the `post_model_hook`! I need to drop for the day but will push a fix & also implement the temp schema name config.

guenp:
All checks are [green](https://github.com/motherduckdb/dbt-duckdb/actions/runs/7744763170)! I also added `tests/functional/plugins/test_motherduck.py` to the MotherDuck tests, I didn't realize your CI/CD ran against a live version of the service so this will help make sure I don't break our product ;)

dbt/adapters/duckdb/impl.py
```diff
@@ -217,6 +230,35 @@ def render_column_constraint(cls, constraint: ColumnLevelConstraint) -> Optional
         else:
             return super().render_column_constraint(constraint)
 
+    def pre_model_hook(self, config: Any) -> None:
+        """A hook for reading"""
```
guenp, line 234: looks like I got distracted while typing this docstring 😁 
```suggestion
        """A hook for reading the model config"""
```

dbt/include/duckdb/macros/materializations/incremental.sql
```diff
@@ -39,13 +41,20 @@
     {% set build_sql = create_table_as(False, intermediate_relation, compiled_code, language) %}
     {% set need_swap = true %}
   {% else %}
+    {% if not temporary %}
+      -- if not using a temporary table we will update the temp relation to use a different temp schema ("dbt_temp" by default)
+      {% set temp_relation = temp_relation.incorporate(path=adapter.get_temp_relation_path(this)) %}
+      {% do run_query(create_schema(temp_relation)) %}
+      -- then drop the temp relation after we insert the incremental data into the target relation
+      {% set need_drop_temp = True %}
```
jwills, line 49: One last nit-- I think this variable is redundant now, right? The `{% if need_drop_temp %}` block below just needs to be `{% if not temporary %}`?

guenp: Great catch!

dbt/adapters/duckdb/environments/__init__.py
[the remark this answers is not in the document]
guenp: Hmm, so I think this fix is now failing in CI and I'm trying to understand why. It does seem to work for me both locally and on the `motherduckdb` fork (see [CI run](https://github.com/motherduckdb/dbt-duckdb/actions/runs/7746950669)) but it's been consistently failing on [this line](https://github.com/duckdb/dbt-duckdb/actions/runs/7746441003/job/21125011980#step:5:351) with the following exception:

```python
>           conn = duckdb.connect(creds.path, read_only=False, config=config)
E           duckdb.duckdb.InvalidInputException: Invalid Input Error: Initialization function "motherduck_init" from file 
"/home/runner/.duckdb/extensions/v0.9.2/linux_amd64_gcc4/motherduck.duckdb_extension" threw an exception: 
"Invalid Input Error: Initialization function "motherduck_init" from file 
"/home/runner/.duckdb/extensions/v0.9.2/linux_amd64_gcc4/motherduck.v1.15.10.duckdb_extension" threw an 
exception: "Unknown Type Error: Cannot connect to MotherDuck server: no token provided.""

/home/runner/work/dbt-duckdb/dbt-duckdb/dbt/adapters/duckdb/environments/__init__.py:142: InvalidInputException
```

I'm not sure why. Either way it seems that both for me locally and for the `motherduckdb` CI, the token is passed via environment variable and the auth to MotherDuck is done upon connection, while in this CI it needs an explicit call to `MD_CONNECT`.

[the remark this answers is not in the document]
jwills: It works fine for me locally as well-- it may simply be that your PR cannot access the GH secret I use for my MD token.

jwills:
Thank you very much @guenp!

guenp:
Yay! 🎉  Could we hold off on pushing a release until @nicku33 has had the chance to test it on a real-life workload? We just want to make sure this workaround is good on our end as well. Thanks!

jwills:
of course, will wait for you all to sign off on it first

guenp:
We had the chance to do a full end to end test with "real life" data so this is signed off on our end!
FYI here's what we tested:
1) using a source from an internal share, make a reasonably sized agg model, something on the order of 1M rows for incremental. Run it once to get a datetime agg for, say, 14 days, then modify it in that incremental kicks in for a good size row set and has to update rows
2) same, but for missing rows. modify the target table so that upon incremental refresh, new rows are filled in
3) same but for deleting rows. modify the source (maybe a copy of a bunch of rows) so that the update produces the correct gaps
4) test the above for the two different incremental styles: `delete + insert` and `append`

PR #327, opened 2024-01-30
added support to dynamically registering/adding partitions

firewall413:
Added support to add new partitions (/locations) to existing or new Glue tables:

Previous PR only registered the Glue table and merely specified its partition columns. 

What if, for each external s3 write, we want to register a new partition in Glue (2024/1/01; 2024/01/02; ...etc).

### DIFF
--- a/dbt/adapters/duckdb/plugins/glue.py
+++ b/dbt/adapters/duckdb/plugins/glue.py
@@ -8,6 +8,7 @@ import boto3
 from mypy_boto3_glue import GlueClient
 from mypy_boto3_glue.type_defs import ColumnTypeDef
 from mypy_boto3_glue.type_defs import GetTableResponseTypeDef
+from mypy_boto3_glue.type_defs import PartitionInputTypeDef
 from mypy_boto3_glue.type_defs import SerDeInfoTypeDef
 from mypy_boto3_glue.type_defs import StorageDescriptorTypeDef
 from mypy_boto3_glue.type_defs import TableInputTypeDef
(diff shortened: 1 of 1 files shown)


dbt/adapters/duckdb/plugins/glue.py
```diff
@@ -132,12 +133,64 @@ def _convert_columns(column_list: Sequence[Column]) -> Sequence["ColumnTypeDef"]
...
+) -> None:
     client.create_table(DatabaseName=database, TableInput=table_def)
+    # Create partition if relevant
+    if partition_columns != []:
+        partition_names = [column["Name"] for column in partition_columns]
+        partition_values = [column["Value"] for column in partition_columns]
 
+        partition_location = table_def["StorageDescriptor"]["Location"]
+        for name, value in zip(partition_names, partition_values):
+            partition_location += f"/{name}={value}"
```
jwills, line 147: I have an aesthetic preference for something like:

```
partition_location = table_def["StorageDescriptor"]["Location"]
partition_components = [partition_location]
for c in partition_columns:
    partition_components.append("=".join((c["Name"], c["Value"]))
partition_location = "/".join(partition_components)
```

dbt/adapters/duckdb/plugins/glue.py
```diff
@@ -132,12 +133,64 @@ def _convert_columns(column_list: Sequence[Column]) -> Sequence["ColumnTypeDef"]
...
+) -> None:
     client.update_table(DatabaseName=database, TableInput=table_def)
+    # Update or create partition if relevant
+    if partition_columns != []:
+        partition_names = [column["Name"] for column in partition_columns]
+        partition_values = [column["Value"] for column in partition_columns]
+
+        partition_location = table_def["StorageDescriptor"]["Location"]
+        for name, value in zip(partition_names, partition_values):
+            partition_location += f"/{name}={value}"
```
jwills, line 170: whatever the logic ends up being for this partition path construction, let's move it to a method so we can unit test it and not repeat it in create/update table

jwills (changes requested):
Ah this is really exciting, thank you so much for adding this; I'd like some type info and code deduplication so we can unit test this stuff; I'm assuming you're using this for work?

dbt/adapters/duckdb/plugins/glue.py
```diff
@@ -132,12 +133,64 @@ def _convert_columns(column_list: Sequence[Column]) -> Sequence["ColumnTypeDef"]
     return column_types
 
 
-def _create_table(client: "GlueClient", database: str, table_def: "TableInputTypeDef") -> None:
+def _create_table(
+    client: "GlueClient", database: str, table_def: "TableInputTypeDef", partition_columns
```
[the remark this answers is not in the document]
firewall413: Done! Decided to stick with List[Dict[str, str]] as I use the partition_columns variable for both creating & updating partitions, I would've preferred: List[ColumnTypeDef] but that doesn't line up with what PartitionInputTypeDef requires. Is that ok?

dbt/adapters/duckdb/plugins/glue.py
firewall413: Done

dbt/adapters/duckdb/plugins/glue.py
firewall413: Done

dbt/adapters/duckdb/plugins/glue.py
```diff
@@ -172,18 +213,35 @@ def _add_partition_columns(table_def: TableInputTypeDef, partition_columns) -> T
...
+        for c in partition_columns:
+            partition_components.append("=".join((c["Name"], c["Value"])))
+        partition_location = "/".join(partition_components)
+
+        partition_input = PartitionInputTypeDef()
+        partition_input["Values"] = partition_values
+        partition_input["StorageDescriptor"] = table_def["StorageDescriptor"]
+        partition_input["StorageDescriptor"]["Location"] = partition_location
+
+    return partition_input, partition_values
```
jwills, line 242: sorry, does this still work if `partition_columns` is empty? I figured it would throw an error since `partition_values` wouldn't be defined.

jwills (commented):
So much better, thank you very much

dbt/adapters/duckdb/plugins/glue.py
firewall413: I'm checking if partition_columns != [] before calling this function as it shouldn't be called with empty partition_colums. Do you prefer that check inside the _parse_partition_columns function?

jwills: Oh I just meant that a default `None` for `partition_values` should be on the same line as the default value for `partition_input`

PR #332, opened 2024-02-06
refactoring-external-materialization

milicevica23:
Hi @jwills,
I wanted to open a pull request to ask questions and potentially propose the refactoring.

I will try in the next few days to list all the requirements and showcase all the implementation of the current external materialization here: https://github.com/l-mds/demo-dbt-duckdb-external-materialization

I would also like to go through that list and implement the missing tests before I start to refactor to ensure that the API stays the same I would also have some questions and need support there.


jwills:
Ah these are all good questions, will inline some comments shortly

dbt/include/duckdb/macros/materializations/external.sql
```diff
@@ -62,10 +67,11 @@
   {{ adapter.rename_relation(intermediate_relation, target_relation) }}
 
   {{ run_hooks(post_hooks, inside_transaction=True) }}
-
+  -- What is should_revoke?
   {% set should_revoke = should_revoke(existing_relation, full_refresh_mode=True) %}
   {% do apply_grants(target_relation, grant_config, should_revoke=should_revoke) %}
 
+  --What does that do?
```
jwills, line 74: This persists table/column comments if they are defined for the relation; DuckDB does not yet have this concept. See also: https://github.com/dbt-labs/dbt-adapters/blob/main/dbt/include/global_project/macros/adapters/persist_docs.sql

dbt/include/duckdb/macros/materializations/external.sql
```diff
@@ -76,9 +82,10 @@
   {{ drop_relation_if_exists(temp_relation) }}
 
   -- register table into glue
+  --I dont know glue so can you explain me a bit about that?
```
jwills, line 85: There's a lot of history here-- the ability to register an external table with the AWS Glue catalog (and thus make it accessible to e.g. AWS Athena) actually predates the concept of dbt-duckdb plugins (in the `BasePlugin` sense.) The original idea of plugins for external relations was to generalize this capability to other cataloging systems.

dbt/include/duckdb/macros/materializations/external.sql
```diff
@@ -76,9 +82,10 @@
...
   {%- set plugin_name = config.get('plugin') -%}
   {%- set glue_register = config.get('glue_register', default=false) -%}
-  {%- set partition_columns = config.get('partition_columns', []) -%}
+  {%- set partition_columns = config.get('partition_columns', []) -%} -- this one is never used?
```
jwills, line 88: Ah, yes-- I think this logic got moved elsewhere and this line was left behind.

dbt/include/duckdb/macros/materializations/external.sql
```diff
@@ -89,6 +96,6 @@
 
   {{ run_hooks(post_hooks, inside_transaction=False) }}
 
-  {{ return({'relations': [target_relation]}) }}
+  {{ return({'relations': [target_relation]}) }} -- to what i return?
```
jwills, line 99: This is returning a `Relation` object in the dbt API: https://github.com/dbt-labs/dbt-adapters/blob/main/dbt/adapters/base/relation.py#L40

dbt/include/duckdb/macros/materializations/external.sql
```diff
@@ -7,11 +7,14 @@
   {%- set read_location = adapter.external_read_location(location, rendered_options) -%}
 
   -- set language - python or sql
+  -- i have to learn general about python models
   {%- set language = model['language'] -%}
 
   {%- set target_relation = this.incorporate(type='view') %}
 
   -- Continue as normal materialization
+  -- Why do we have temp and intermediate relation?
```
jwills, line 16: The idea is to not destroy an existing relation that has the same name as this one (if it exists in the catalog) in case something fails during the construction of the relation defined by this model.

dbt/include/duckdb/macros/materializations/external.sql
```diff
@@ -7,11 +7,14 @@
...
 
   -- set language - python or sql
+  -- i have to learn general about python models
   {%- set language = model['language'] -%}
 
   {%- set target_relation = this.incorporate(type='view') %}
 
   -- Continue as normal materialization
+  -- Why do we have temp and intermediate relation? 
+  -- What does load_cached_relation do?
```
jwills, line 17: Built-in dbt method; dbt will do the equivalent of caching the DB catalog metadata it loaded during the run so as to minimize the number of round-trips it does to the database (less relevant for DuckDB, more relevant for e.g. MotherDuck.)

milicevica23:
Hi @jwills, 
Thank you. I will try to add my comments and further questions and, in the next few days, make a first poc of how I think this should work. Please be critical

The main goal, for now, is to showcase the code structure, then we can go into details. 

To simplify a bit, I will move out from the scope for now:
* glue registration -> I don't have enough experience with the glue, so I have to check it out and will come to that afterward (it should work before merging)
* arrow batch streaming -> so the first version will be with the arrow Table, which is the exact materialization, rather than exporting in the batches. see https://twitter.com/milicevica23/status/1754802477575078041 (important for delta)
* assume that I don't reference a model in the different run so that functionality from "register_upstream_external_models" is not needed for now (it should work before merging)

dbt/include/duckdb/macros/materializations/external.sql
```diff
@@ -27,6 +30,8 @@
   {%- set backup_relation = make_backup_relation(target_relation, backup_relation_type) -%}
   -- as above, the backup_relation should not already exist
   {%- set preexisting_backup_relation = load_cached_relation(backup_relation) -%}
+  
+  --What is grants here?
```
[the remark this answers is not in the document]
milicevica23: I will skip it for now to simplify, but the access grants here are an exciting topic. We want to export from the current context/duckdb into external parquet, delta, iceberg, and register a view over this location so that the table can be referenced in the next steps.

dbt/include/duckdb/macros/materializations/external.sql
```diff
@@ -62,10 +67,11 @@
   {{ adapter.rename_relation(intermediate_relation, target_relation) }}
 
   {{ run_hooks(post_hooks, inside_transaction=True) }}
-
+  -- What is should_revoke?
```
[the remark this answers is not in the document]
milicevica23: understand, but can't imagine for now

dbt/include/duckdb/macros/materializations/external.sql
milicevica23: aha, it pushes the information from the yaml files into the dwh and this is per adapter, i will also skip this one for the simplicity for now but will come back to that

dbt/include/duckdb/macros/materializations/external.sql
milicevica23: Nice, I see; I will need you to support me from the requirements point of view because I am unsure if I can imagine all possible use cases implemented until now. I don't have so much experience in the AWS world so more azure, google cloud but happy to learn about glue and athena
I will skip this one till we don't agree on some structure of the code; then, we can speak about details and each use case.

dbt/include/duckdb/macros/materializations/external.sql
milicevica23: i understand; we prepare everything and do all the transformation, and at the very end, we exchange the names

I understand how it fits into the context but am not sure what it means for me right now in the implementation, so make sure that i don't overlook that one in the implementation.

dbt/include/duckdb/macros/materializations/external.sql
milicevica23: I will have to take a look into MotherDuck here because I still don't understand how this fits here and what the places are where I have to be careful not to overlook the functionality. So please, if you see something that is against MotherDuck logic, make me aware of that.

milicevica23:
So, I added the structure and will start with the simple implementation of the native plugin, which will hopefully show if this makes sense.

This is still a work in progress, but If you have some comments or questions I would be happy to answer them.

jwills:
> So, I added the structure and will start with the simple implementation of the native plugin, which will hopefully show if this makes sense.
> 
> This is still a work in progress, but If you have some comments or questions I would be happy to answer them.
> 
> I have an general question. Is there some concept of the release candidate / nightly that people can try this before they start to use it :D i am not sure if i can get all 1-to-1 api in the first run

There isn't a nightly or some such thing as of yet; if we add it, we will add it as a net-new thing that is disabled by default (which is how we've done a bunch of other things, like e.g. the fast CSV loading for `dbt seed`)

milicevica23:
Hi @jwills, can you take a look. I did the first draft of the native functionality and a new test which should cover some of the general cases

milicevica23:
Hi @jwills,
I think I finished the first round of refactoring, so maybe you could take a look and see if this makes sense to you.
What I did: 
- refactored the external materialization so that the creation of the external tables is handled by the plugin
- the native plugin is a new plugin which does the parquet, json, csv
- adapted the Excel and sqlalchemy plugins to use new code flow.
- tried to fulfill and make all test run that were in place without big change
- adapter upstream macro to register needed tables and data frames

The new code flow provides a nice feature that the exported file can be pointed directly from the target destination. It sounds trivial for native format but it is especially important for the further implementations of the iceberg, delta, (big query, snowflake ?) adapters, which can now be implemented within external materializations. 
It also provides a very nice convenient way to register df in the upstream macro

What is there still to do:
missing:
- Glue integration -> currently, i skipped it but will take a look and would maybe need your help for this because I never used it 
- python part of the external materialization -> i don't know so much about this and didn't find tests with that so I am not sure how to go about this but will take a look
- buenavista integration -> what is the status there? i have to see what should be supported and adapt to that 

has to do: 
- native plugin has to be initialized on its own as default, and I have to find a place where to do it 
- add more tests for excel, sqlalchemy and others -> we have to take use cases from the community in order to cover all cases 
- look into the upstream macro because i think that with the old implementation, not all cases were done correctly because there is the problem with the rendering -> will try to explain in the future but the https://stackoverflow.com/questions/22655031/jinja2-how-to-put-a-block-in-an-if-statement shows how the rendering works and i think that we didn't use it right

nice to have: 
- maybe add some tests with azurite and minio to simulate object storage

I hope this gives you a bit overview of what i did, i am happy to answer all the questions

I still fight with the mypy and formater in the pre-commit if you have some suggestions how to go about this i would appreciate it a lot and of

jwills:
@milicevica23 ah that is very cool, thanks for the update! I have some stuff going on this week but I will dive into this on Thursday!

jwills:
@milicevica23 when you get a chance could you redo the formatting according to the `pre-commit run --all-files` setup? All of the formatting diffs make this hard to review

dbt/adapters/duckdb/plugins/__init__.py
```diff
@@ -121,9 +123,17 @@ def load(self, source_config: SourceConfig):
...
         raise NotImplementedError(f"load method not implemented for {self.name}")
 
-    def store(self, target_config: TargetConfig):
+    # coursor is needed just for the native, we have to do it better
+    # to had it over in some initalization?
+    def store(self, df: DuckDBPyRelation, target_config: TargetConfig, cursor=None):
         raise NotImplementedError(f"store method not implemented for {self.name}")
 
+    def create_source_config(self, target_config: TargetConfig) -> SourceConfig:
+        raise NotImplementedError(f"store method not implemented for {self.name}")
```
jwills, line 132: copy-pasta

dbt/adapters/duckdb/plugins/__init__.py
```diff
@@ -110,7 +111,8 @@ def configure_connection(self, conn: DuckDBPyConnection):
         """
         pass
 
-    def load(self, source_config: SourceConfig):
+    # coursor is needed for the native plugin
+    def load(self, source_config: SourceConfig, coursor=None):
```
jwills, line 115: nit: cursor

dbt/adapters/duckdb/plugins/native.py
```diff
@@ -0,0 +1,108 @@
...
+            database=target_config.relation.database,
+            meta=target_config.as_dict(),
+            tags=[],
+        )
+        return source_config
+
+    def store(self, df: DuckDBPyRelation, target_config: TargetConfig, cursor=None):
+        location = target_config.location.path
+        options = external_write_options(location, target_config.config.get("options", {}))
+        cursor.sql(f"COPY (SELECT * FROM df) to '{location}' ({options})")
```
jwills, line 52: Yeah so this is my core question: does this extra-level of indirection here (query to df to query of DF that writes to an external file) introduce overhead or a perf hit that is going to cause issues for existing `external` users? Or is the DuckDB optimizer smart enough to optimize this away (for e.g. arrow record batches or some such thing?)

dbt/adapters/duckdb/plugins/postgres.py
```diff
@@ -6,6 +6,8 @@
 from . import BasePlugin
 
 
+## This is maybe what we can deprecate? Native way in the config
+# is better or there is some need?
 class Plugin(BasePlugin):
```
jwills, line 11: Separate issue, but yes we can almost certainly deprecate this plugin shortly

dbt/adapters/duckdb/plugins/sqlalchemy.py
```diff
@@ -2,40 +2,60 @@
...
 from sqlalchemy import create_engine
 from sqlalchemy import text
 
 from . import BasePlugin
-from . import pd_utils
 from ..utils import SourceConfig
 from ..utils import TargetConfig
 
+# overall i would recommend to use attach in duckdb but this can be nice for some cases where
+# native one is not supported e.g starrock
```
jwills, line 14: agreed

jwills (commented):
Some notes/comments/Qs-- general comment that there is lots of "coursor" vs. "cursor" stuff in the codebase now, would be good to do a find/replace on "coursor" -> "cursor"

Biggest Q is the one on the write perf impact of DuckDB doing an extra level of indirection via arrow before it writes the results of a query to a file

dbt/adapters/duckdb/plugins/native.py
milicevica23: Hi @jwills, a very good question
I am unsure if here goes and convert to arrow, but I would empirically try something today to showcase if this is the case. I intentionally hand over `DuckDBPyRelation ` so the plugin `store` function can choose the needed conversion.
Independent of the result, we would have a `target_config.config.model.compiled_code` variable, which we can embed into the string, and this should be the same as in the old path.

milicevica23: Hi @jwills, in my unscientific empirical experiment, we can see there that the new flow/method is faster than the previous one, where we first create a table and then export it
I also added the conversion over the arrow and pandas data frame to make it better comparable 

image

The code is here: https://github.com/milicevica23/wtf-memory/blob/main/002-basics-duckdb/export-parquet-benchmark.py

Please be aware that there can be some mistakes in terms of cache and swapping memory, but I think it is enough to show that there is no significant difference in the concept of how duckdb retrieves the data

milicevica23: I also did a bit of experimenting with memory usage to see if i can visualize it; the b variant is better because there is no intermediate materialization 

version b)
image

version c)

dbt/adapters/duckdb/plugins/__init__.py
milicevica23: done

dbt/adapters/duckdb/plugins/__init__.py
milicevica23: done

myfsuhail:
Is there any update on the PR for the support of delta/iceberg/hudi table format? 

This would be a great feature!! Let me know if there is an estimate and plan for the support!! 

Thanks in advance!!

jwills:
I'm mostly opting to defer to the support for Iceberg/Delta that are getting built into DuckDB extensions (e.g. https://github.com/duckdb/uc_catalog indicates where I think things are going); I don't want to have a custom way of handling these table formats/catalogs inside of dbt-duckdb that is different from the extensions unless it's absolutely necessary.

milicevica23:
As @jwills we kinda stopped here because of the native support and it is better to focus on the underlying functionality which is benefiting for more users from duckdb than just dbt-duckdb users
@jwills I think we can close my two pull requests?

PR #335, opened 2024-02-14
Fix: MotherDuck token from plugin config should override env var

guenp:
This PR fixes a bug:
- When passing a token via the config, this didn't work because there is a typo (by yours truly 😬) [here](https://github.com/duckdb/dbt-duckdb/blob/8a6e0dc33cfedd580be1b0a0a8e970b6fce05a5f/dbt/adapters/duckdb/plugins/motherduck.py#L16).

The PR also addresses how a user can pass the MotherDuck token to `duckdb.connect`:

- Adds a `motherduck_token` property to `DuckDBCredentials` that fetches it from the plugin config.
- The user can specify the token in the path (e.g. `"md:?motherduck_token=<token>"`. This will override a `motherduck_token` environment variable if it exists.
- We no longer use an explicit `SET` command and instead pass a `config` dictionary to `duckdb.connect`. Any config options passed this way will override the corresponding environment variable as well as any options set in the path.

So the behavior is now the same as in duckdb: env var < path < config. (where left < right means "right overrides left").

### DIFF
--- a/dbt/adapters/duckdb/credentials.py
+++ b/dbt/adapters/duckdb/credentials.py
@@ -165,7 +165,7 @@ class DuckDBCredentials(Credentials):
         """
         plugins = self.plugins or []
         for plugin in plugins:
-            if plugin.module == "motherduck" and plugin.config:
+            if plugin.config:
                 token = plugin.config.get("token") or ""
                 return str(token)
         return ""
--- a/dbt/adapters/duckdb/credentials.py
+++ b/dbt/adapters/duckdb/credentials.py
@@ -154,7 +154,7 @@ class DuckDBCredentials(Credentials):
     @staticmethod
     def _is_motherduck(scheme: str) -> bool:
         return scheme in {"md", "motherduck"}
-    
+
     @property
     def motherduck_token(self) -> str:
         plugins = self.plugins or []
--- a/dbt/adapters/duckdb/credentials.py
+++ b/dbt/adapters/duckdb/credentials.py
@@ -156,11 +156,12 @@ class DuckDBCredentials(Credentials):
         return scheme in {"md", "motherduck"}
 
     @property
-    def motherduck_token(self) -> str:
+    def motherduck_token(self) -> Optional[str]:
         plugins = self.plugins or []
         for plugin in plugins:
-            if plugin.module == "motherduck":
-                return plugin.config.get("token")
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -40,9 +40,7 @@ def _dbt_duckdb_version():
         "dbt-core~=1.7.0",
         "duckdb>=0.7.0",
     ],
-    extras_require={
-        "glue": ["boto3", "mypy-boto3-glue"],
-    },
+    extras_require={"glue": ["boto3", "mypy-boto3-glue"], "md": ["duckdb>=0.7.0,<=0.9.2"]},
```
guenp, line 43: I added this so the CI succeeds (MotherDuck currently does not support 0.10).
See run here: https://github.com/motherduckdb/dbt-duckdb/actions/runs/7895023998/job/21546711483

jwills: No worries

dbt/adapters/duckdb/credentials.py
```diff
@@ -155,6 +155,15 @@ def is_motherduck(self):
     def _is_motherduck(scheme: str) -> bool:
         return scheme in {"md", "motherduck"}
 
+    @property
+    def motherduck_token(self) -> str:
```
jwills, line 159: So I'm a little thrown by this-- I thought the normal place folks put their MD token was in the `path` setting as opposed to by using the MD plugin configuration. There may be more places we want to look for the token besides the plugin config?

guenp: I moved it out of this file entirely :)

guenp:
All checks are green: https://github.com/motherduckdb/dbt-duckdb/actions/runs/7909652885/job/21591671925

After we merge this we should be OK to release! 🎉

setup.py
jwills, line 43: Do you want me to keep this bit in for now?

jwills (commented):
Looks great @guenp -- just one Q and then I will merge it

setup.py
guenp: Yes please! I think will will be super helpful to have `pip install dbt-duckdb[md]` as an option ready to go for people to use before we roll out 0.10 support, and I can open a PR to bump this version once that lands

guenp: (We also need to keep this line in tox.ini)
```
-e.[md]
```

PR #355, opened 2024-03-06
Add automated release pipeline

guenp:
See [packaging guide](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) for setup instructions on PyPI:

> Go to https://pypi.org/manage/account/publishing/.
>
> Fill in the name you wish to publish your new [PyPI project](https://packaging.python.org/en/latest/glossary/#term-Project) under (the name value in your setup.cfg or pyproject.toml), the GitHub repository owner’s name (org or user), and repository name, and the name of the release workflow file under the .github/ folder, see [Creating a workflow definition](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#workflow-definition). Finally, add the name of the GitHub Environment (pypi) we’re going set up under your repository. Register the trusted publisher.
>
>Your “pending” publisher is now ready for their first use and will create your project automatically once you use them for the first time.

This was tested on TestPyPI (see: https://test.pypi.org/project/dbt-duckdb/)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,9 @@
-name: Publish Python 🐍 distribution 📦 to PyPI and TestPyPI
+name: Publish Python 🐍 distribution 📦 to PyPI
 
-on: push
+on:
+  push:
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+'
 
 jobs:
   build:
@@ -91,27 +94,3 @@ jobs:
         gh release upload
         '${{ github.ref_name }}' dist/**
         --repo '${{ github.repository }}'
-
-  publish-to-testpypi:
-    name: Publish Python 🐍 distribution 📦 to TestPyPI
-    needs:
-    - build
-    runs-on: ubuntu-latest
-
-    environment:
-      name: testpypi
-      url: https://test.pypi.org/p/dbt-duckdb
-
-    permissions:
-      id-token: write  # IMPORTANT: mandatory for trusted publishing
-
-    steps:
-    - name: Download all the dists
-      uses: actions/download-artifact@v3
-      with:
-        name: python-package-distributions
-        path: dist/
-    - name: Publish distribution 📦 to TestPyPI
-      uses: pypa/gh-action-pypi-publish@release/v1
-      with:
-        repository-url: https://test.pypi.org/legacy/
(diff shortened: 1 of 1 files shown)


guenp:
The pipeline currently triggers when you push a semver tag but still requires you to update the version manually, we can implement automated versioning e.g. with `versioningit` in a separate PR. Let me know what you think

.github/workflows/release.yml
```diff
@@ -0,0 +1,96 @@
...
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
+      run: >-
+        gh release create
+        '${{ github.ref_name }}'
+        --repo '${{ github.repository }}'
+        --notes ""
+    - name: Upload artifact signatures to GitHub Release
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
```
guenp, line 89: Ah I missed this earlier, looks like if we want to push an automatic release as well we need to set up a github token.

jwills: okay cool-- so I need to set that up, somehow. It's int he docs you linked to?

guenp: Oh good point, no it's not. I'll open a PR to add a note.

I looked it up and looks like the token itself is automatically generated by GitHub, see [docs](https://docs.github.com/en/actions/security-guides/automatic-token-authentication).
However we would need to set the permissions so it has `contents:write` access, see [docs](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release).
See also [this post](https://stackoverflow.com/questions/67389957/what-permissions-does-github-token-require-for-releases-from-a-github-action) on stackoverflow.

To change the permissions go to  https://github.com/duckdb/dbt-duckdb/settings/actions and scroll down to "Workflow permissions" and choose "Read and write permissions".

Alternatively you can create a [Personal Access Token](https://docs.github.com/en/enterprise-server@2.22/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) and use that instead of the `GITHUB_TOKEN`.

If none of this sparks joy we can take this part out and just manually post release notes

guenp: Also just realized that right now it posts empty release notes (`--notes ""`)
Maybe a nice follow up PR could be to add automated release notes?
See: https://github.com/marketplace/actions/release-notes-generator#use-github-action

PR #364, opened 2024-03-17
Switch the docs from meta -> config as the default way of spec'ing source settings

jwills:
The `config` settings won't propagate into `dbt docs generate`, which is what you want when you're including secrets in the config, so make it the default here


jwills:
Need to hack around this: https://github.com/dbt-labs/dbt-core/issues/9759

mycaule:
Some users might find interesting though to keep the s3:// or abfss:// paths for documentation.

jwills:
Going to replace this with more detailed docs on the options here

PR #378, opened 2024-04-13
Only run MotherDuck integration tests on master, skip for pull requests

guenp:
This PR adds a job that checks if the MotherDuck token is defined, and only runs the integration tests if it is. This way the job is just skipped instead of failing with a `Cannot connect to MotherDuck server: no token provided.` error.
Additionally, I've updated the `push` trigger branch name to `master` so the tests always run after PRs are merged.

For more detailed info see:

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
https://gist.github.com/GuillaumeFalourd/d3524cf6361afad87b9da3bf78270bc1

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ name: Tests and Code Checks
 on:
   push:
     branches:
-      - "main"
+      - "master"
       - "develop"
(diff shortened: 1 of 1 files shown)


jwills:
Thank you @guenp !

PR #381, opened 2024-04-17
postgres_scanner plugin

obronco:


obronco:
https://duckdb.org/2022/09/30/postgres-scanner.html

jwills:
Looks like this one got conflated with your s3 profile one-- let's rebase it and take another look? And was the intent here just to switch the postgres plugin over to use the postgres_scanner extension in DuckDB?

obronco:
Sorry about the S3 stuff. Regarding postgres_scanner, I added it for personal use because of performance issues I was having with the default postgres plugin. I thought it would help others.

jwills:
Closing this out in favor of a more recent PR for the postgres plugin

PR #382, opened 2024-04-17
Added profile_name to boto3 Session

obronco:
Needed when working with multiple S3 profiles on the same machine


jwills:
Ah this seems like a good idea, will merge this and then take a look at the linter stuff-- thanks!

PR #386, opened 2024-04-22
Fix flaky attach test

jwills:
I did a bad job of writing this test and so it's flaky on GH Actions-- asked Claude to fix it for me and this is what he came up with


jwills:
Ah hah-- this is a real regression against 0.10.2 instead of 0.10.1, not a flaky test, my bad-- will see if I can fix it

PR #391, opened 2024-05-03
Bump duckdb version to 0.10.2 for MotherDuck integration tests

guenp:
This bumps the DuckDB version to 0.10.2 for MotherDuck integration tests. I've added a new secret `MOTHERDUCK_TOKEN_10` that I set up specifically for this pipeline that we can use here and is on the MD 0.10.x preview. I kept the `MOTHERDUCK_TOKEN` for now in case we need to revert back to that for some reason.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -208,7 +208,7 @@ jobs:
     steps:
         - id: md-token
           env:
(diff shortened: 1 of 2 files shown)


guenp:
Closed this because I couldn't get it to access the new secret from the fork.
New PR open here: https://github.com/duckdb/dbt-duckdb/pull/392

PR #392, opened 2024-05-03
Bump md duckdb version to 0.10.2 and use new token

guenp:
This bumps the DuckDB version to 0.10.2 for MotherDuck integration tests. I've added a new secret `MOTHERDUCK_TOKEN_10` that I set up specifically for this pipeline that we can use here and is on the MD 0.10.x preview. I kept the `MOTHERDUCK_TOKEN` for now in case we need to revert back to that for some reason.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -201,14 +201,14 @@ jobs:
           path: filebased_results.csv
 
   check-md-token:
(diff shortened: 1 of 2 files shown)


jwills:
@guenp good to merge this for the last 1.7.x release?

PR #399, opened 2024-05-30
Pin to duckdb v0.10.2 to avoid view dependencies

guenp:
This PR adds a constraint to setup.py to skip `duckdb` v0.10.3 and avoid issues with view dependency enforcement. This will be fixed in 0.10.4 and above by the `enable_view_dependencies` configuration setting (which is set to false by default).

Related issue: https://github.com/duckdb/duckdb/issues/12191

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ setup(
     install_requires=[
         "dbt-common>=1,<2",
         "dbt-adapters>=1,<2",
-        "duckdb!=0.10.3",
+        "duckdb>=0.7.0,!=0.10.3",
         # add dbt-core to ensure backwards compatibility of installation, this is not a functional dependency
         "dbt-core>=1.8.0",
     ],
--- a/setup.py
+++ b/setup.py
@@ -39,11 +39,11 @@ setup(
     install_requires=[
         "dbt-common>=1,<2",
         "dbt-adapters>=1,<2",
-        "duckdb>=0.7.0",
+        "duckdb!=0.10.3",
         # add dbt-core to ensure backwards compatibility of installation, this is not a functional dependency
         "dbt-core>=1.8.0",
     ],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -39,11 +39,11 @@ def _dbt_duckdb_version():
     install_requires=[
         "dbt-common>=1,<2",
         "dbt-adapters>=1,<2",
-        "duckdb>=0.7.0",
+        "duckdb!=0.10.3",
```
jwills, line 42: I thought it was like `duckdb>=0.7.0,!=0.10.3`?

jwills (commented):
Just a question on how this works-- I don't normally have this problem

setup.py
guenp: Yeah you're right, I forgot that `duckdb<0.7.0` is no bueno

setup.py
```diff
@@ -39,11 +39,11 @@ def _dbt_duckdb_version():
     install_requires=[
         "dbt-common>=1,<2",
         "dbt-adapters>=1,<2",
-        "duckdb>=0.7.0",
+        "duckdb>=0.7.0,!=0.10.3",
         # add dbt-core to ensure backwards compatibility of installation, this is not a functional dependency
         "dbt-core>=1.8.0",
     ],
-    extras_require={"glue": ["boto3", "mypy-boto3-glue"], "md": ["duckdb>=0.10.2"]},
+    extras_require={"glue": ["boto3", "mypy-boto3-glue"], "md": ["duckdb==0.10.2"]},
```
edgarrmondragon, line 46: I think this change isn't needed? The constraint on the top-level dependency should prevent 0.10.3 from being installed if the user requests `dbt-duckdb[md]`.

jwills: Maybe not, but I don't think it does any harm

guenp: This is only used by the CI right now

PR #406, opened 2024-06-19
Add DATE to dateadd macro

### LINKED ISSUE #405: Duck Db support of DATEADD from Core
I tried the example here: 
https://docs.getdbt.com/sql-reference/dateadd
```
SELECT {{ dateadd(datepart="month", interval=1, from_date_or_timestamp="'2021-08-12'") }}
```
This fails. 

I tweaked and this works: 
```
SELECT {{ dateadd(datepart="month", interval=1, from_date_or_timestamp="DATE '2021-08-12'") }}  AS period_of_load
```

In general I think adapters should support the documented case from core. 

Here duckdb seems to want date strings to be explicitly refererred to as dates in addition operations as per: https://duckdb.org/docs/sql/functions/date.html

As such I suggest that : 
https://github.com/duckdb/dbt-duckdb/blob/master/dbt/include/duckdb/macros/utils/dateadd.sql

Is updated to include: DATE at start

alittlesliceoftom:
Fixes #405 

tl;dr, helps you avoid errors like: 

```
  Binder Error: Could not choose a best candidate function for the function call "+(STRING_LITERAL, INTERVAL)". In order to select one, please add explicit type casts.        Candidate functions:
        +(DATE, INTERVAL) -> TIMESTAMP
        +(TIME, INTERVAL) -> TIME
        +(TIMESTAMP, INTERVAL) -> TIMESTAMP
        +(TIME WITH TIME ZONE, INTERVAL) -> TIME WITH TIME ZONE
        +(TIMESTAMP WITH TIME ZONE, INTERVAL) -> TIMESTAMP WITH TIME ZONE
        +(INTERVAL, INTERVAL) -> INTERVAL

  LINE 1: /* {"app": "dbt", "dbt_version": "1.7.9", "profile_...

```

### DIFF
--- a/dbt/include/duckdb/macros/utils/dateadd.sql
+++ b/dbt/include/duckdb/macros/utils/dateadd.sql
@@ -1,5 +1,5 @@
 {% macro duckdb__dateadd(datepart, interval, from_date_or_timestamp) %}
 
-    date_add({{ from_date_or_timestamp }}, interval ({{ interval }}) {{ datepart }})
+    date_add(DATE {{ from_date_or_timestamp }}, interval ({{ interval }}) {{ datepart }})
 
 {% endmacro %}


alittlesliceoftom:
See docs here: https://duckdb.org/docs/sql/functions/date.html

alittlesliceoftom:
I didn't have time to setup integration tests, but tested the updated macro locally, and 'correct' example now works: 


 SELECT 
    {{ dateadd(datepart="month", interval=1, from_date_or_timestamp="'2021-08-12'") }}  AS period_of_load

jwills:
yeah I'm surprised this didn't break on the regular unit tests-- like, does this not work anymore and I missed it? https://github.com/dbt-labs/dbt-adapters/blob/main/dbt-tests-adapter/dbt/tests/adapter/utils/fixture_dateadd.py

alittlesliceoftom:
Possibly those are more explicit strings?

jwills:
My concern here is what happens if I pass in a `TIMESTAMP` instead of a string or a `DATE`? Or a string that should be cast as a `TIMESTAMP` instead?

alittlesliceoftom:
Yeh... Interestingly in DuckDb the dateadd operations seem to expect the coder to know and set the type ahead of time in the statement? 
https://duckdb.org/docs/sql/functions/interval

jwills:
Closing this out as I believe the code was working the way dbt intended

PR #418, opened 2024-07-23
Prep for 1.8.2 release

guenp:
`dbt-core` is on 1.8.4 and [other adapters](https://docs.getdbt.com/docs/trusted-adapters) are all still using 1.8.x or below versioning. However, since the recent [refactor](https://github.com/dbt-labs/dbt-adapters), `dbt` adapters no longer require staying in pace with `dbt` versioning. From `dbt` team:
> We generally recommend that major version bumps reflect major changes to the adapter and should therefore be reserved. Minor versions should reflect meaningful changes to behavior.

This release includes a major version bump for DuckDB from 0.10.x to 1.0.0 and support for the new DuckDB Secrets Manager, which could be categorized as a "meaningful changes" to the behavior (the feature gap between 0.10.x and 1.0.0 isn't huge - it was mostly bugfixes and stability improvements - and the secrets behavior is an addition that is backward compatible).

I started a discussion on `dbt` community Slack here: [#adapter-ecosystem](https://getdbt.slack.com/archives/C030A0UF5LM).

Update: Bumping **major** version based on discussion with `dbt` team
Update: Sticking with **minor** version for now - once `dbt-core` 1.9 comes out, we're not sure if 1.8.x adapters will remain compatible, and bumping the version again will cause more confusion
Update: OK, after a few more chats with the `dbt` adapters team and evaluating pros and cons, we decided to go with a minor release for now. Let it be so 🪄

### DIFF
--- a/dbt/adapters/duckdb/__version__.py
+++ b/dbt/adapters/duckdb/__version__.py
@@ -1,1 +1,1 @@
-version = "1.8.1"
+version = "1.8.2"
--- a/dbt/adapters/duckdb/__version__.py
+++ b/dbt/adapters/duckdb/__version__.py
@@ -1,1 +1,1 @@
-version = "1.8.2"
+version = "1.9.0"
--- a/dbt/adapters/duckdb/__version__.py
+++ b/dbt/adapters/duckdb/__version__.py
@@ -1,1 +1,1 @@
-version = "1.9.0"
+version = "1.8.2"


jwills:
Pro case: this _is_ a major change (the secrets manager switchover and the DuckDB 1.0.0 constraint).

Con case: will be confusing to users, I'm not 100% confident that the dbt folks are actually going to maintain compat with 1.8.x adapters when they do dbt-core 1.9.0 and will we then need to do a dbt-duckdb 1.9.x that fixes the compatibility issues with dbt-core 1.9.x which will be extra-extra confusing

guenp:
That makes sense. For now we can just keep the adapter major version synced with the with `dbt-core` major version, and after they go through the 1.9 update we can reevaluate. wdyt?

guenp:
(I meant to just comment but fat fingered the wrong button)

PR #421, opened 2024-07-31
fix(secrets): value should be enclosed in quotes

waaronmorris:
[Resolving Issue 420](https://github.com/duckdb/dbt-duckdb/issues/420)

### DIFF
--- a/dbt/adapters/duckdb/secrets.py
+++ b/dbt/adapters/duckdb/secrets.py
@@ -47,7 +47,7 @@ class Secret(dbtClassMixin):
         params.update(params.pop("secret_kwargs", {}))
         params_sql = f",\n{tab}".join(
             [
-                f"{key} '{value}'"
+                f"{key} '{value}'" if key not in ["type", "provider"] else f"{key} {value}"
                 for key, value in params.items()
                 if value is not None and key not in ["name", "persistent"]
             ]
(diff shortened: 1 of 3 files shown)


jwills:
ack, thank you @waaronmorris! cc @guenp

guenp:
Thanks @waaronmorris, this is a great catch!
Could you also update the [unit tests](https://github.com/duckdb/dbt-duckdb/blob/ec65e430bf387235924805a75585dda6ce3ede67/tests/unit/test_credentials.py#L39)?

dbt/adapters/duckdb/secrets.py
```diff
@@ -47,7 +47,7 @@ def to_sql(self) -> str:
         params.update(params.pop("secret_kwargs", {}))
         params_sql = f",\n{tab}".join(
             [
-                f"{key} {value}"
+                f"{key} '{value}'"
```
guenp, line 50:
```suggestion
                f"{key} '{value}'" if key not in ["type", "provider"] else f"{key} {value}"
```

guenp (changes requested):
We also need to add quotes to all the unit tests - see https://github.com/duckdb/dbt-duckdb/commit/3190554bcd926946e7e548e4be4828d64b6cfe55

PR #427, opened 2024-08-16
(chore) Use pbr for packaging for automated versioning with git tag, move to setup.cfg

guenp:
Using some tools to (hopefully) make our lives easier :)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,7 +28,6 @@ install_requires=
     duckdb>=1.0.0
     # add dbt-core to ensure backwards compatibility of installation, this is not a functional dependency
     dbt-core>=1.8.0
-    pbr>=6.0.0
 python_requires = >=3.8
 include_package_data = True
 packages = find_namespace:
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -0,0 +1,51 @@
...
+    distutils
+
+[options]
+install_requires=
+    dbt-common>=1,<2
+    dbt-adapters>=1,<2
+    duckdb>=1.0.0
+    # add dbt-core to ensure backwards compatibility of installation, this is not a functional dependency
+    dbt-core>=1.8.0
+    pbr>=6.0.0
```
jwills, line 31: Wait is this right? The user of the package doesn't need to install `pbr`, right? It's just a dev dependency?

guenp: Ah good point
I was using `pbr` to get the package version in `version.py` but I can just get that with `importlib` instead. will update

setup.cfg
```diff
@@ -0,0 +1,50 @@
+[metadata]
+name = dbt-duckdb
+author = Josh Wills
+author-email = joshwills+dbt@gmail.com
```
SaravananSathyanandhaQC, line 4: @guenp @jwills should this be `author_email`? I think `author-email` is deprecated

jwills: https://github.com/duckdb/dbt-duckdb/pull/428

PR #435, opened 2024-09-05
Disable Python models on MotherDuck when SaaS mode is on

guenp:


jwills:
LGTM, fsspec test is a rate limit thing

PR #444, opened 2024-09-20
(fix) Add support for attaching a MotherDuck database with token in settings or config

guenp:
Add support for attaching a MotherDuck database and pass an arbitrary token via `config`, `settings` or `path`.

Example profiles.yml:
```
test:
  outputs:
    user:
      threads: 1
      type: duckdb
      path: /tmp/data.duckdb
      attach:
       - path: md:my_db
         type: motherduck
      settings:
        motherduck_token: "{{ env_var('motherduck_token_1') }}"
```
Right now, this fails with an error:
```
Runtime Error
  setting 'motherduck_token' can only be set during initialization
```
This is because the `SET` command happens after the `ATTACH`. Similarly, when passing the token via `config`, this also fails, because the `motherduck` extension is not yet loaded when running `duckdb.connect`. Lastly, passing the token via the `path` also fails because it's passed to the `ATTACH` command directly and this also does not let you set the token.

Therefore, currently the only way to attach a MotherDuck database is to have the `motherduck_token` env var set.

After some discussion with Josh we decided:

* re-order the attach block and the plugin initialization block inside of initialize_db (so that plugins were initialized before attaches)
* `SET` the MotherDuck token before the md: attach happens. Similar to how the config is updated when connecting to a MD path.


dbt/adapters/duckdb/plugins/motherduck.py
```diff
@@ -8,40 +10,73 @@
...
 
-        # If a user specified the token via the plugin config,
-        # pass it to the config kwarg in duckdb.connect
-        token = self.token_from_config(creds)
-        if token != "":
-            config["motherduck_token"] = token
+        # If a user specified MotherDuck config options via the plugin config,
+        # pass it to the config kwarg in duckdb.connect.
+        if not creds.is_motherduck_attach:
+            if self._token:
```
edgarrmondragon, line 81: I think this can be simplified to

```suggestion
        if not creds.is_motherduck_attach and self._token:
```

jwills (approved):
looks good to me @guenp!

PR #446, opened 2024-09-26
(fix): MotherDuck config should set SaaS mode at the end

guenp:
- make sure to set `saas_mode` _after_ the MD token is set
- generalize how config settings are parsed from input
- add tests to confirm that SaaS mode also works when I attach a MD database
- pin DuckDB to 1.1.1 for MD tests


.github/workflows/main.yml
```diff
@@ -222,7 +222,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.9']
+        python-version: ['3.11']
```
jwills, line 225: Doing a separate PR to update all of these now that 3.8 is basically at EOL

tests/functional/plugins/motherduck/test_motherduck_attach.py
```diff
@@ -93,17 +93,17 @@ def run_dbt_scope(self, project, database_name):
         # CREATE DATABASE does not work with SaaS mode on duckdb 1.0.0
         # This will be fixed in duckdb 1.1.0
```
jwills, line 94: Really 1.1.1?

tests/functional/plugins/motherduck/test_motherduck_saas_mode.py
```diff
@@ -89,16 +89,97 @@ def run_dbt_scope(self, project, database_name):
         # CREATE DATABASE does not work with SaaS mode on duckdb 1.0.0
         # This will be fixed in duckdb 1.1.0
```
jwills, line 90: same

.github/workflows/main.yml
guenp: Cool! I've removed this commit

tests/functional/plugins/motherduck/test_motherduck_saas_mode.py
guenp: indeed!

tests/functional/plugins/motherduck/test_motherduck_attach.py
guenp: This actually doesn't need to go here so I removed it!

jwills (approved):

PR #449, opened 2024-10-02
Test with dbt-core prereleases

edgarrmondragon:


jwills:
@edgarrmondragon ack-- thanks for these fixes!

jwills (approved):

PR #450, opened 2024-10-03
Add support for configuring community/nightly extensions in the profile directly

jwills:
Wanted to have proper support for community/nightly extensions from dbt-duckdb for folks available in the `profiles.yml` without requiring pre-hooks to set them up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,9 +59,28 @@ option that will be automatically enabled if you are connecting to a MotherDuck 
 
 #### DuckDB Extensions, Settings, and Filesystems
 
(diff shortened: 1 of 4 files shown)


guenp:
Looks like the `logger` errors is because we have to change
```python
from dbt.logger import GLOBAL_LOGGER as logger
```

into

```python
from dbt.adapters.events.logging import AdapterLogger

logger = AdapterLogger("duckdb")
```

see https://github.com/dbt-labs/dbt-adapters/tree/main/dbt/adapters/events#adapter-maintainers

guenp (approved):

PR #453, opened 2024-10-04
(fix) generate_database_name macro should be case insensitive

guenp:
Fixes https://github.com/duckdb/dbt-duckdb/discussions/376

### DIFF
--- a/dbt/include/duckdb/macros/adapters.sql
+++ b/dbt/include/duckdb/macros/adapters.sql
@@ -40,8 +40,8 @@
   {% set sql -%}
         select count(*)
         from system.information_schema.schemata
-        where schema_name = '{{ schema }}'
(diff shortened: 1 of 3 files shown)


dbt/include/duckdb/macros/adapters.sql
```diff
@@ -157,7 +157,7 @@ def materialize(df, con):
         END as type
     from system.information_schema.tables
     where table_schema = '{{ schema_relation.schema }}'
```
jwills, line 159: macro question (so to speak): do we have the same issue for schemas w/respect to cases and what not? Do we need to do the same casing stuff for schemas as well?

guenp: yeah why not. I'd be OK with extending this to `generate_schema_name` as well

jwills (approved):

PR #463, opened 2024-10-15
adding transaction support in md

matsonj:
Cleaning up readme in preparation for adding this documentation to the [dbt docs](https://docs.getdbt.com/docs/introduction).


matsonj:
Wondering aloud (+@guenp) if we should format these docs to match the dbt page exactly for ease of maintenance. I've re-organized slightly on the dbt side but I am not sure if I should.

jwills (approved):
rubberstamping

matsonj:
> Wondering aloud (+@guenp) if we should format these docs to match the dbt page exactly for ease of maintenance. I've re-organized slightly on the dbt side but I am not sure if I should.

On second thought, these are aiming for different goals. Will need to keep them slightly differently, although only just.

PR #464, opened 2024-10-16
Added basic AWS Glue struct datatype support and JSON/CSV/PARQUET read_options

firewall413:
Added basic AWS Glue struct datatype support and JSON/CSV/PARQUET read_options

### DIFF
--- a/dbt/include/duckdb/macros/materializations/external.sql
+++ b/dbt/include/duckdb/macros/materializations/external.sql
@@ -5,6 +5,8 @@
   {%- set format = config.get('format', 'parquet') -%}
   {%- set write_options = adapter.external_write_options(location, rendered_options) -%}
   {%- set read_location = adapter.external_read_location(location, rendered_options) -%}
+  {%- set parquet_read_options = config.get('parquet_read_options', {'union_by_name': False}) -%}
+  {%- set json_read_options = config.get('json_read_options', {'maximum_object_size': 1000000000}) -%}
 
   -- set language - python or sql
   {%- set language = model['language'] -%}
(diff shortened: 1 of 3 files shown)


jwills:
Amazing, thank you @firewall413! Looking forward to reading this one

tests/functional/adapter/test_external.py
```diff
@@ -23,11 +23,11 @@
 """
 
 config_materialized_parquet_location = """
-  {{ config(materialized="external", location="{{ adapter.external_root() }}/test.parquet") }}
+  {{ config(materialized="external", location="{{ adapter.external_root() }}/test.parquet", format="parquet") }}
```
jwills, line 26: Ah so this is a behavior change? i.e., you now have to specify the `format` arg for external materializations in order to get the read options right for the corresponding view?

jwills: Given that we need this info to config the views correctly, can we just infer it from the extension on the file if the format isn't specified? (And if there is no extension at all  and no format specified, I think the default is parquet)?

Context here is that I don't want to break existing users when they upgrade unless it is completely unavoidable and even then we want to provide as much warning that it is coming as possible.

firewall413: ok yep, I can give that (infer the type from the extension) a try!

dbt/include/duckdb/macros/materializations/external.sql
```diff
@@ -2,9 +2,22 @@
...
   {%- set rendered_options = render_write_options(config) -%}
-  {%- set format = config.get('format', 'parquet') -%}
+
+  {%- set format = config.get('format') -%}
+  {%- set allowed_formats = ['csv', 'parquet', 'json'] -%}
+
+  {%- if format -%}
+      {%- set format = format if format in allowed_formats else 'parquet' -%}
+  {%- else -%}
+      {%- set format = location.split('.')[-1] if '.' in location else 'parquet' -%}
```
jwills, line 12: we might also want to do a lower on the format string here so we don't fail for `.CSV` or `.JSON` (or worse, silently write out parquet instead of the requested format)

dbt/include/duckdb/macros/materializations/external.sql
```diff
@@ -45,13 +58,55 @@
...
+        {%- if value is string -%}
+          '{{ value }}'
+        {%- else -%}
+          {{ value }}
+        {%- endif -%}
+      {%- endfor -%}
+      )
     );
+    {% else %}
+      create or replace view {{ intermediate_relation }} as (
```
jwills, line 106: This condition is impossible right now, right? No objection to leaving this block in, just wanted to be sure I had this right

jwills (approved):
Yay, thank you again @firewall413!

PR #470, opened 2024-10-22
Remove mention of outdated MotherDuck constraints in Readme

Flogex:
MotherDuck is now backwards-compatible with old client DuckDB versions. Also, foreign keys and macros are supported. Therefore, two of the constraints mentioned when using dbt with MotherDuck can be removed from the readme.


jwills (approved):

jwills:
Hey, sorry I missed this-- thank you!

PR #472, opened 2024-10-29
Add safety check for node in graph

tehunter:
With dbt's unit-test feature, dbt may not populate the graph when running `dbt test -s model_to_test`. This will result in an error in `register_upstream_external_models`:

```
on-run-start failed, error:
 'dict object' has no attribute 'unit_test.project_name.model_name.test_name'
```

This change checks if a node is present before looking for its dependencies.


jwills (approved):
Thank you @tehunter!

PR #476, opened 2024-11-16
Document Postgres attachment for read/write

dumkydewilde:
Currently the documentation only shows using Postgres as a plugin. This uses the `POSTGRES_SCANNER` implementation of DuckDB, however using the [Postgres extension](https://duckdb.org/docs/extensions/postgres.html) as an attachment allows for both reading and writing. Though the current implementation of attachments in DuckDB seems to be intended for file-based attachments, this functionality works for postgres to using a Postgres DSN. 

Potentially additional implementation changes can be made to either improve the attachment functionality or even remove the Postgres plugin in favour of the attachment implementation, but I'll leave that up to a discussion. For now I think a simple documentation change can save people a lot of headaches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,13 +163,15 @@ default:
           read_only: true
         - path: sqlite.db
           type: sqlite
+        - path: postgresql://username@hostname/dbname
+          type: postgres
 ```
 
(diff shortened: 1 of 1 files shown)


jwills (approved):
Looks good-- thank you!

PR #493, opened 2024-12-17
add support for per_thread_output in external materialization

elephantmetropolis:
When using the DuckDB `COPY ... TO` command with the `per_thread_output` parameter, it can create multiple files, similar to the behavior of the `partition_by` parameter. These parameters can be handled in almost the same way.

Possible solution to that issue I opened https://github.com/duckdb/dbt-duckdb/issues/492


jwills (approved):
LGTM-- thank you @elephantmetropolis!

PR #497, opened 2025-01-09
Run tests with DuckDB nightly release every 24h

guenp:
Add nightly tox env that installs duckdb nightly release, add GitHub action to periodically trigger nightly tests


jwills (approved):
Thank you for this @guenp , should help a lot!

PR #502, opened 2025-01-21
Make keep_open: true the default setting for dbt-duckdb going forward

jwills:
Recent versions of DuckDB (>= 1.13) seem to perform better/more reliably when we leave the connection to the DuckDB file open for the entire DuckDB run vs. the old behavior of having different threads open/close the file in rapid succession, so moving to a model where we generally take ownership of the DuckDB file for the length of any dbt run unless explictly told not to in the profiles.yml.


jwills:
@guenp that plugins test is failing for delta tables for seemingly unrelated reasons (i tested it w/ and w/o the keep_open settings in place)

guenp (approved):

PR #507, opened 2025-02-09
Update deprecated v3 -> v4 for upload-artifact and download-artifact

jwills:
Just some housekeeping here per https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/


jwills:
I'm going to pin pyarrow to 18.1.0 until this is fixed (https://github.com/apache/arrow/issues/45283 )-- was hoping it would be sooner, alas

guenp (approved):

PR #510, opened 2025-02-10
[Fix] External materialization to S3 does not work if table is empty

guenp:
Workaround for https://github.com/duckdb/dbt-duckdb/issues/503 (need to look into fix in `duckdb` for writing empty tables to S3)

It consists of inserting `NULL` values (if the table is empty), writing those to S3, and then creating a view that filters those `NULL` values


jwills (approved):
LGTM

PR #520, opened 2025-03-14
Vibe-claudeing a duckdbt CLI

jwills:
Been playing with the new DuckDB UI stuff and Claude Code a lot this week and cooked up something simple that does what I've always wanted to be able to do with dbt and DuckDB: run the dbt-duckdb process in a shell process where I could simultaneously query the DuckDB instance via a UI and run my dbt-duckdb models in a high-fidelity way, with all of the python model execution and all of the other great stuff we've built in this project over the last few years.

The idea is to create a `cli.py` and a new `duckdbt` shell command that uses the [dbtRunner](https://docs.getdbt.com/reference/programmatic-invocations) construct to load up dbt, parse all of the models in a project using a given profile, and let you run command invocations in the CLI while simultaneously launching the DuckDB UI via a `CALL start_ui()` call against the instance of DuckDB that gets loaded and configured when `dbt debug` is called.

### DIFF
--- a/dbt/adapters/duckdb/cli.py
+++ b/dbt/adapters/duckdb/cli.py
@@ -3,10 +3,12 @@ import cmd
 import os
 import shlex
 import sys
-from typing import List, Optional
+from typing import List
+from typing import Optional
 
 from dbt.adapters.duckdb.connections import DuckDBConnectionManager
-from dbt.cli.main import dbtRunner, dbtRunnerResult
+from dbt.cli.main import dbtRunner
+from dbt.cli.main import dbtRunnerResult
 
 
 class DuckdbtShell(cmd.Cmd):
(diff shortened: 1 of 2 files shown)


guenp (approved):
image

This is pretty cool

b-per:
It worked super well, thanks!
The only thing is that we then lose the [shell completion](https://github.com/dbt-labs/dbt-completion.bash) capabilities for models and integrations like https://github.com/Infused-Insight/fzf-dbt

dbeatty10:
For anyone that wants to try this out, I did the following steps within a Python virtual environment.

First, `cd` to a directory on your filesystem that has a dbt project that has a `dbt-duckdb` connection profile.

If you don't already have one, here's an easy and fast way to get one:
```shell
git clone https://github.com/dbt-labs/jaffle_shop_duckdb.git
cd jaffle_shop_duckdb
```

Then install `dbt-duckdb` (and dependencies) from source:

```shell
python -m pip install git+https://github.com/duckdb/dbt-duckdb.git@master
```

> [!NOTE]  
> Once the `duckdbt` CLI is included in the [latest release of `dbt-duckdb`](https://pypi.org/project/dbt-duckdb/#history), then it will be just this instead of installing from source:
> ```shell
> python -m pip install --upgrade dbt-duckdb
> ```


Then start using this CLI (which will open a window in your default web browser containing the [DuckDB UI](https://duckdb.org/2025/03/12/duckdb-ui.html)):

```shell
duckdbt
```

You'll probably need to rearrange your windows so that browser window is stacked on top of your terminal window (or however else you like it).

Then in the terminal you can run your typical dbt sub commands (just drop the `dbt ` prefix):
```shell
build
list
build -s stg_customers
```

And in the DuckDB UI window you can run queries:
```sql
select * from stg_customers
```

jwills:
Ah @dbeatty10 , nice to see you! Checkout https://github.com/duckdb/dbt-duckdb/pull/524 !

PR #523, opened 2025-03-20
Fix/register partitioned tables in glue

firewall413:
- resolved issue when writing a new partition to glue for a table with the same schema would result in multiple glue schema versions (instead of just adding a partition)
- resolved issue where when writing new partition to Glue: The schema of the file at **external_read_location */*/*.parquet** would be picked up instead of actual last written partition **year=2024/month=03/day=20/*.parquet**


jwills:
@firewall413, hey nice to see you-- and thanks for putting this together!

firewall413:
Still working on those failing tests, my bad, I'll finish up after my holidays!

PR #524, opened 2025-03-22
Add fzf-based model autocomplete to duckdbt

jwills:
Had Claude take a crack at this based on feedback from @b-per ; the `duckdbt` CLI will now auto-complete models using fzf if `iterfzf` is installed in the python environment.

(We update the cache of models in the project based on the manifest whenever `parse` is run)


guenp (approved):
thanks vibe Claude!
looks good to me

PR #529, opened 2025-03-31
Support dbt_valid_to_current for snapshots

### LINKED ISSUE #528: Handle dbt_valid_to_current in snapshots
Starting with `dbt 1.9`, there's a possibility to  configure `dbt_valid_to_current` in [snapshots](https://docs.getdbt.com/reference/resource-configs/dbt_valid_to_current), such that `valid_to` is not null. 

When `dbt_valid_to_current` is configured, `dbt-duckdb` fails to update the previous records with the update date.

### Steps to reproduce

- create an ephemeral model, `stg_test_snapshot_ephemeral`
```
{{ config(materialized='ephemeral') }}

select range as id,
       now() at time zone 'UTC' as updated_at
from range(2)
```
- define the snapshot, under `snapshots`:

```
  - name: test_snapshot
    relation: ref('stg_test_snapshot_ephemeral')
    config:
      unique_key: id
      updated_at: updated_at
      strategy: timestamp
      dbt_valid_to_current: "'9999-12-31 23:59:59'::timestamp"
```

After two executions (`dbt snapshot --select test_snapshot`), the previous records remain open:
```
select id, dbt_updated_at, dbt_valid_to from test_snapshot;
┌───────┬─────────────────────────┬─────────────────────┐
│  id   │     dbt_updated_at      │    dbt_valid_to     │
│ int64 │        timestamp        │      timestamp      │
├───────┼─────────────────────────┼─────────────────────┤
│     0 │ 2025-03-31 08:39:53.729 │ 9999-12-31 23:59:59 │
│     1 │ 2025-03-31 08:39:53.729 │ 9999-12-31 23:59:59 │
│     0 │ 2025-03-31 08:39:57.712 │ 9999-12-31 23:59:59 │
│     1 │ 2025-03-31 08:39:57.712 │ 9999-12-31 23:59:59 │
└───────┴─────────────────────────┴─────────────────────┘
```

Looking at the logs, the calculation of dbt_valid_to takes into account the default value

```
  coalesce(nullif(updated_at, updated_at), '9999-12-31'::date) as dbt_valid_to
```

, but the update on the target table filters on `DBT_INTERNAL_TARGET.dbt_valid_to is null;`.

jwills:
Fixes #528

### DIFF
--- a/dbt/include/duckdb/macros/snapshot_helper.sql
+++ b/dbt/include/duckdb/macros/snapshot_helper.sql
@@ -1,12 +1,18 @@
 {% macro duckdb__snapshot_merge_sql(target, source, insert_cols) -%}
     {%- set insert_cols_csv = insert_cols | join(', ') -%}
 
+    {%- set columns = config.get("snapshot_table_column_names") or get_snapshot_table_column_names() -%}
+
     update {{ target }} as DBT_INTERNAL_TARGET
-    set dbt_valid_to = DBT_INTERNAL_SOURCE.dbt_valid_to
+    set {{ columns.dbt_valid_to }} = DBT_INTERNAL_SOURCE.{{ columns.dbt_valid_to }}
     from {{ source }} as DBT_INTERNAL_SOURCE
-    where DBT_INTERNAL_SOURCE.dbt_scd_id::text = DBT_INTERNAL_TARGET.dbt_scd_id::text
+    where DBT_INTERNAL_SOURCE.{{ columns.dbt_scd_id }}::text = DBT_INTERNAL_TARGET.{{ columns.dbt_scd_id }}::text
       and DBT_INTERNAL_SOURCE.dbt_change_type::text in ('update'::text, 'delete'::text)
-      and DBT_INTERNAL_TARGET.dbt_valid_to is null;
+      {% if config.get("dbt_valid_to_current") %}
+        and (DBT_INTERNAL_TARGET.{{ columns.dbt_valid_to }} = {{ config.get('dbt_valid_to_current') }} or DBT_INTERNAL_TARGET.{{ columns.dbt_valid_to }} is null);
+      {% else %}
+        and DBT_INTERNAL_TARGET.{{ columns.dbt_valid_to }} is null;
+      {% endif %}
 
     insert into {{ target }} ({{ insert_cols_csv }})
     select {% for column in insert_cols -%}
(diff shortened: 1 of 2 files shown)


dbt/include/duckdb/macros/snapshot_helper.sql
```diff
@@ -1,12 +1,18 @@
...
     update {{ target }} as DBT_INTERNAL_TARGET
-    set dbt_valid_to = DBT_INTERNAL_SOURCE.dbt_valid_to
+    set {{ columns.dbt_valid_to }} = DBT_INTERNAL_SOURCE.{{ columns.dbt_valid_to }}
     from {{ source }} as DBT_INTERNAL_SOURCE
-    where DBT_INTERNAL_SOURCE.dbt_scd_id::text = DBT_INTERNAL_TARGET.dbt_scd_id::text
+    where DBT_INTERNAL_SOURCE.{{ columns.dbt_scd_id }}::text = DBT_INTERNAL_TARGET.{{ columns.dbt_scd_id }}::text
       and DBT_INTERNAL_SOURCE.dbt_change_type::text in ('update'::text, 'delete'::text)
-      and DBT_INTERNAL_TARGET.dbt_valid_to is null;
+      {% if config.get("dbt_valid_to_current") %}
+        and (DBT_INTERNAL_TARGET.{{ columns.dbt_valid_to }} = {{ config.get('dbt_valid_to_current') }} or DBT_INTERNAL_TARGET.{{ columns.dbt_valid_to }} is null);
```
guenp, line 12: Do we need to cast these to `TIMESTAMP`?
From the issue it seems doing the cast in the yaml also works but might not hurt to do it here as well in case people pass just a string:
```
  - name: test_snapshot
    relation: ref('stg_test_snapshot_ephemeral')
    config:
      unique_key: id
      updated_at: updated_at
      strategy: timestamp
      dbt_valid_to_current: "'9999-12-31 23:59:59'::timestamp"
```

guenp (approved):

dbt/include/duckdb/macros/snapshot_helper.sql
jwills: oh interesting, and I don't know the answer; the code I used for this change was a DuckDB-ified version of the code from dbt-postgres here: https://github.com/dbt-labs/dbt-postgres/blob/main/dbt/include/postgres/macros/materializations/snapshot_merge.sql#L12

@dbeatty10 would be grateful if you have a sec to answer @guenp's question. 🙇

dbeatty10: Good question!

[`dbt_valid_to_current`](https://docs.getdbt.com/reference/resource-configs/dbt_valid_to_current) is expected to be a YAML `string` representing a SQL expression that returns a SQL date/time data type (like `timestamp`) in the target database.

e.g., something like one of these (depending on the database):

```sql
to_date('9999-12-31')
date(YYYY, MM, DD)
'9999-12-31 23:59:59'::timestamp
```

etc.

To the best of my knowledge, a simple string literal like the following examples is not officially supported for `dbt_valid_to_current` in `dbt-postgres`:
```sql
'9999-12-31'
9999-12-31 23:59:59
```

While we didn't add that cast to `dbt-postgres`, it's your prerogative if you want to add that proposed cast or not for `dbt-duckdb`.

PR #530, opened 2025-04-01
Add support for scoped secrets/credentials by storage prefix

01100100:
This PR introduces small changes to support `scope` in duckdb secrets created through a duckdb profile eg) `profile.yml`.

Context: I wanted to query data in different s3 buckets, they have different regions and require different credentials. This was somewhat supported before, however multiple "scopes" per secret where not supported. It was is possible with duckdb secrets manager, but I had to create the secrets manually.

This PR to create the secrets through a duckdb profile. For the below example, the first secret will be used anytime the first two buckets path is queried and the second secret will be used for the other.

```yaml
default:
  outputs:
    dev:
      type: duckdb
      path: /tmp/dbt.duckdb
      extensions:
        - httpfs
        - parquet
      secrets:
        - type: s3
          provider: credential_chain
          scope: [ "s3://bucket-in-eu-region", "s3://bucket-2-in-eu-region" ]
          region: "eu-central-1"
        - type: s3
          region: us-west-2
          scope: "s3://bucket-in-us-region"
```

Reference: [https://duckdb.org/docs/stable/configuration/secrets_manager.html](https://duckdb.org/docs/stable/configuration/secrets_manager.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,31 @@ default:
   target: dev
 ```
 
+#### Scoped credentials by storage prefix
+
+Secrets can be scoped, such that different storage path can use different credentials.
+
+```
+default:
+  outputs:
+    dev:
+      type: duckdb
+      path: /tmp/dbt.duckdb
+      extensions:
+        - httpfs
+        - parquet
+      secrets:
+        - type: s3
+          provider: credential_chain
+          scope: [ "s3://bucket-in-eu-region", "s3://bucket-2-in-eu-region" ]
+          region: "eu-central-1"
+        - type: s3
+          region: us-west-2
+          scope: "s3://bucket-in-us-region"
+```
+
+When fetching a secret for a path, the secret scopes are compared to the path, returning the matching secret for the path. In the case of multiple matching secrets, the longest prefix is chosen.
+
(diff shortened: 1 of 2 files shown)


jwills:
This looks good to me, but it looks like it broke some of the tests we had here since we're changing how the credential SQL gets generated-- would you mind fixing them up @01100100 ? 🙇

01100100:
@jwills I updated the test, seems to run locally :smile:

guenp:
@01100100 thank you this is great!
If it's not too much trouble would you mind adding a unit test for this that shows what the SQL is that should get generated when providing multiple scopes for a secret? You can copy one of the existing tests and modify it.

01100100:
@guenp, ofcourse, added a test with multiple scopes and amended the commit here. 

I ran tox locally and everything passes, also checked the expected sql is valid in the duck 🦆 🍇  Let me know if there is anything else 😄 

Thanks for maintaining this project, I am a big fan.

guenp (approved):
This is super neat, thank you! 🙌

01100100:
Very much welcome 🦆

Issue #533, opened 2025-04-03
[bug] delete + insert not wrapped in a transaction


rchui:
I don't have the most concrete evidence for it, but I've encountered a couple of occurrences that lead me to believe that this is the case. This happens very rarely, but there are times where a table will spontaneously have double the number of records that it should have immediately following a full refresh of a table.

What I believe is happening is if I have a `--full-refresh` on a table running and then I initiate an incremental on the same table, there appears to be a race condition where at the moment that the full refresh tables are swapped, the incremental inserts into the freshly swapped table as part of it's delete + insert strategy.

I couldn't find the specific place this would be occurring but this has happened too many times to me to be a coincidence.

jwills:
Huh-- this is on a MotherDuck table? (I assume it has to be b/c I don't think you could do this on a local DuckDB file?)

rchui:
@jwills yes this is on a MotherDuck database

jwills:
Ah, then I suspect I know what the problem is-- MD used to not support transactions at all, so dbt-duckdb disables them by default on dbt-duckdb connections. You can disable this disabling by putting "disable_transactions: false" into your dbt profile, but I suspect it is time to remove that old default. /cc @guenp 

Will prep a PR for it shortly

rchui:
@jwills thank you! Where in the dbt profile would that go? Under settings?

jwills:
@rchui yes, on the same level of the yaml as the `type` and `path` settings (i.e., `disable_transactions` is a top-level dbt-duckdb setting)

Issue #532, opened 2025-04-03
[bug] tables are not swapped in transactions


rchui:
It appears that table full refresh `--full-refresh` does not swap tables within a transaction. In rare instances we are able to observe the table missing but the `__dbt_tmp` table is present:

```
Catalog Error: Table with name ${table} does not exist!
Did you mean "${table}__dbt_tmp"?

LINE 92:       FROM ${database}.${schema}.${table}
```

Missing tables create downtime when data is being read off of the tables at the same time that the tables are being refreshed. They should be wrapped in transactions to ensure that this downtime does not affect readers.

Ref: https://duckdb.org/docs/stable/sql/statements/transactions.html

guenp:
Just making sure I follow -- @rchui , is the `__dbt_tmp` mentioned here because of a concurrent incremental run?
Is this related to/a dupe of #533 ?

rchui:
@guenp this is related to #533 .

The order of full refresh table replace is roughly:
1. `alter table {table} rename to {table}__dbt_backup`
2. `alter table {table}__dbt_tmp rename to {table}`
3. `drop table {table}__dbt_backup`

Both stem from the fact that transactions are not enabled by default in MotherDuck. The read scaling instances are transactionally consistent but the table swap / rename / drop when full refreshing a model is not. This error appears when the read scaling instances snapshot between the two table renames [1, 2] due to them not being wrapped in a transaction.

guenp:
@rchui I don't fully understand the sequence of events. Where are you seeing this error? On a read scaling instance or on the read/write instance?

I just opened https://github.com/duckdb/dbt-duckdb/pull/540 to re-enable txns for MotherDuck by default.

rchui:
@guenp this error is visible on both the read scaling and read/write instances. The table swap is non-transactional so the swap is visible if you query either of the instances at just the right time.

guenp:
OK great, I'll close this then since we merged #540. Thanks!

PR #534, opened 2025-04-03
MotherDuck supports transactions now, so we don't need to disable them by default anymore

jwills:
This probably fixes a whole host of issues, but most directly #533 and #532

### DIFF
--- a/dbt/adapters/duckdb/credentials.py
+++ b/dbt/adapters/duckdb/credentials.py
@@ -228,11 +228,7 @@ class DuckDBCredentials(Credentials):
             parsed = urlparse(path)
             base_file = os.path.basename(parsed.path)
             path_db = os.path.splitext(base_file)[0]
-            # For MotherDuck, turn on disable_transactions unless
-            # it's explicitly set already by the user
             if cls._is_motherduck(parsed.scheme):
-                if "disable_transactions" not in data:
-                    data["disable_transactions"] = True
                 if path_db == "":
                     path_db = "my_db"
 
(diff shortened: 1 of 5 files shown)


guenp:
thanks @jwills ! I don't think we can drop the temp table workaround just yet, we still need those to enable incremental models to run on MotherDuck.

jwills:
> thanks @jwills ! I don't think we can drop the temp table workaround just yet, we still need those to enable incremental models to run on MotherDuck.

Yeah I'm v confused b/c when I left the temp table thing in, the incremental models for the bad unique key test failed: https://github.com/duckdb/dbt-duckdb/actions/runs/14255394805/job/39957333057 for reasons that I couldn't quite suss out; does that error make sense to you? (I was also able to repro it locally using my MD account)

guenp:
@jwills can we hold off on this change for now? We have a customer who ran into a WAL corruption after they enabled transactions, and I want to do some more testing before we enable this by default for MotherDuck users. I can send you a new PR in a few days and also look into why the temp tables test is failing.

jwills:
yes of course!

guenp:
Closing this for now, thanks again 🙇‍♂️

guenp:
@jwills we are g2g to turn txns back on for MotherDuck, I reopened a PR
the reason why the incremental test was failing is because of the `duckdb` version bump, since v1.2.0 error messages are truncated which causes [this assertion](https://github.com/dbt-labs/dbt-adapters/blob/main/dbt-tests-adapter/src/dbt/tests/adapter/incremental/test_incremental_unique_id.py#L489) to fail -- we didn't catch this in our nightly builds because those run only against local duckdb and this error is unique to MotherDuck's `dbt_temp` workaround
I'll see if there is a way to not truncate the error msg and open a separate PR to bump duckdb to the latest version

PR #540, opened 2025-04-21
Fix: turn txns back on for MD

### LINKED ISSUE #533: [bug] delete + insert not wrapped in a transaction

guenp:
Reopening #534 
Fixes #533

### DIFF
--- a/dbt/adapters/duckdb/credentials.py
+++ b/dbt/adapters/duckdb/credentials.py
@@ -228,11 +228,7 @@ class DuckDBCredentials(Credentials):
             parsed = urlparse(path)
             base_file = os.path.basename(parsed.path)
             path_db = os.path.splitext(base_file)[0]
-            # For MotherDuck, turn on disable_transactions unless
-            # it's explicitly set already by the user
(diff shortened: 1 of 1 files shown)


jwills:
Do we still need the dbt_temp etc. workaround with this change, or can that go too?

guenp:
> Do we still need the dbt_temp etc. workaround with this change, or can that go too?

we still need that unfortunately, right now temp tables get created locally by default and that causes an extra roundtrip from MD to the local machine which slows down incremental runs.

PR #544, opened 2025-04-23
Add SQLGlot Plugin to transpile from any SQL dialect into DuckDb

HuyNguyen7994:
One of the most common use case for DuckDb is to run SQL tests locally. [SQLMesh](https://github.com/TobikoData/sqlmesh) supports this natively with [SQLGlot](https://github.com/tobymao/sqlglot), but no alternative exists for dbt workflow. There's global macros but they don't cover many use cases. In my case it's to transpile `to_date` in postgres into `strptime` in duckdb. This plugin facilitates that by intercepting the original compiled sql (either in redshift, postgres, or mysql), and transpiling them into duckdb syntax.

### DIFF
--- a/dbt/adapters/duckdb/environments/__init__.py
+++ b/dbt/adapters/duckdb/environments/__init__.py
@@ -229,6 +229,9 @@ class Environment(abc.ABC):
         for df_name, df in registered_df.items():
             cursor.register(df_name, df)
 
+        if "sqlglot" in plugins:
+            cursor = plugins["sqlglot"].cursor(cursor)
+
         if creds.retries and creds.retries.query_attempts:
             cursor = RetryableCursor(
                 cursor, creds.retries.query_attempts, creds.retries.retryable_exceptions
--- a/dbt/adapters/duckdb/plugins/sqlglot.py
+++ b/dbt/adapters/duckdb/plugins/sqlglot.py
@@ -1,0 +1,35 @@
+from typing import Any
+from typing import Dict
+
+import sqlglot
+from dbt_common.exceptions import DbtRuntimeError
+
+from . import BasePlugin
+
+
+class TranspiledCursor:
+    def __init__(self, cursor, trans_from: str):
+        self.trans_from = trans_from
+        self._cursor = cursor
+
+    # forward along all non-execute() methods/attribute look-ups
+    def __getattr__(self, name):
+        return getattr(self._cursor, name)
+
+    def execute(self, sql, bindings=None):
+        sql = sqlglot.transpile(sql, read=self.trans_from, write="duckdb").pop()
+        try:
+            if bindings is None:
+                return self._cursor.execute(sql)
+            else:
+                return self._cursor.execute(sql, bindings)
+        except RuntimeError as e:
+            raise DbtRuntimeError(str(e))
+
+
+class Plugin(BasePlugin):
+    def initialize(self, plugin_config: Dict[str, Any]):
+        self.trans_from = plugin_config.get("trans_from", "duckdb")
+
+    def cursor(self, cursor):
+        return TranspiledCursor(cursor, self.trans_from)
--- a/tests/functional/plugins/test_sqlglot.py
+++ b/tests/functional/plugins/test_sqlglot.py
@@ -1,0 +1,54 @@
+import pytest
+
+from dbt.adapters.duckdb.plugins.postgres import Plugin
+from dbt.tests.fixtures.project import TestProjInfo
+from dbt.tests.util import check_relations_equal, run_dbt
+
+
+class TestSqlglotPlugin:
+    @pytest.fixture(scope="class")
+    def profiles_config_update(self, dbt_profile_target):
+        return {
+            "test": {
+                "outputs": {
+                    "dev": {
+                        "type": "duckdb",
+                        "path": dbt_profile_target.get("path", ":memory:"),
+                        "plugins": [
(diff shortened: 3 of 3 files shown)


jwills:
Ah, nice to see someone taking another crack at this one @HuyNguyen7994 -- I've never been confident that this would work well enough in the general case to include it in the project, although maybe a plugin-based approach is the right way to experiment with it? See https://github.com/duckdb/dbt-duckdb/pull/22 for my prior attempts to do this.

Did you have a dbt project where you used sqlglot to do this transpiling and had it work well?

HuyNguyen7994:
No, this is still also experimental on my part. In my case it's to transpile between different flavors of Postgres syntax, so my problem set is much smaller.

HuyNguyen7994:
Somehow changing where the cursor is modified fix all the errors ...

calswbin:
@HuyNguyen7994 are there any updates with this one?

goobill:
Can I suggest changing this function slightly.

```python

from dbt.adapters.events.logging import AdapterLogger

logger = AdapterLogger("DuckDB:SQLGlot")

class SqlglotWrapper(DuckDBCursorWrapper):
    def __init__(self, cursor, sql_from: str):
        self.sql_from = sql_from
        self._cursor = cursor

    def execute(self, sql, bindings=None):
        try:
            transpiled = transpile(
                sql, 
                read=self.sql_from, 
                write="duckdb", 
                identify=True, 
                pretty=False, 
                comments=False, 
                unsupported_level=errors.ErrorLevel.IMMEDIATE,
                error_level=errors.ErrorLevel.IMMEDIATE
            )
            sql = ";".join(transpiled)
        except errors.SqlglotError as err:
            logger.warning(err)

        super().execute(sql, bindings)
```

If the SQL contains multiple statements i.e. INSERT; SELECT; the transpile command will return multiple SQL statements.

---

---

---

I also found running this with Databricks SQL with DBT incremental load wouldn't work. 

DBT creates it's own queries for the incremental load but wraps the columns with double quotes. When this is transpiled these double quotes get converted to single quotes. This then causes a run time error on DuckDB.

```
14:25:45    Runtime Error in model stg_events (models/example/stg_events.sql)
  Parser Error: syntax error at or near "'event_time'"
  
  LINE 1: INSERT INTO dev.main.stg_events ('event_time') (SELECT 'event_time' FROM stg_events__dbt_tmp...
```

For example

```sql
INSERT INTO dev.main.stg_events ("event_time") (SELECT "event_time" FROM stg_events__dbt_tmp);
SELECT 'foo bar bob' AS example_text;
INSERT INTO Customers ("CustomerName", "ContactName", "Address", "City", "PostalCode", "Country")
VALUES ('Cardinal', 'Tom B. Erichsen', 'Skagen 21', 'Stavanger', '4006', 'Norway');
```

Would be transpiled to

```sql
INSERT INTO dev.main.stg_events ('event_time') (SELECT 'event_time' FROM stg_events__dbt_tmp);
SELECT 'foo bar bob' AS example_text;
INSERT INTO Customers ('CustomerName', 'ContactName', 'Address', 'City', 'PostalCode', 'Country') 
VALUES ('Cardinal', 'Tom B. Erichsen', 'Skagen 21', 'Stavanger', '4006', 'Norway')
```

I have a temporary fix

```python
def unquote_columns(expression: exp.Expression) -> exp.Expression:
    for node in expression.walk():
        if isinstance(node, (exp.Schema, exp.Select)) and node.expressions:
            for lit in node.expressions:
                if isinstance(lit, exp.Literal) and lit.args.get("is_string"):
                    lit.set("is_string", False)
    return expression

expressions = parse(sql, dialect=self.sql_from)
transformed = [unquote_columns(expr) for expr in expressions]
sql = ";".join(expr.sql(dialect=self.sql_from) for expr in transformed)
```

This focuses the literals in Schema() and Select() objects.

```
[Insert(
   this=Schema(
     this=Table(
       this=Identifier(this=stg_events, quoted=False),
       db=Identifier(this=main, quoted=False),
       catalog=Identifier(this=dev, quoted=False)),
     expressions=[
       Literal(this='event_time', is_string=True)]),
   expression=Subquery(
     this=Select(
       expressions=[
         Literal(this='event_time', is_string=True)],
       from=From(
         this=Table(
           this=Identifier(this=stg_events__dbt_tmp, quoted=False)))))),
 Select(
   expressions=[
     Alias(
       this=Literal(this='foo bar bob', is_string=True),
       alias=Identifier(this=example_text, quoted=False))]),
 Insert(
   this=Schema(
     this=Table(
       this=Identifier(this=Customers, quoted=False)),
     expressions=[
       Literal(this='CustomerName', is_string=True),
       Literal(this='ContactName', is_string=True),
       Literal(this='Address', is_string=True),
       Literal(this='City', is_string=True),
       Literal(this='PostalCode', is_string=True),
       Literal(this='Country', is_string=True)]),
   expression=Values(
     expressions=[
       Tuple(
         expressions=[
           Literal(this='Cardinal', is_string=True),
           Literal(this='Tom B. Erichsen', is_string=True),
           Literal(this='Skagen 21', is_string=True),
           Literal(this='Stavanger', is_string=True),
           Literal(this='4006', is_string=True),
           Literal(this='Norway', is_string=True)])]))]
```

Not sure this is the correct solution as it seems a bit hacky. But it seems to be a problem with DBT and transpiling to DuckDB.

Ill add it to my own local version of this plugin but curious if anyone else found this a problem and if there was a better way of going about solving it.

PR #548, opened 2025-05-02
Apply project quoting setting to add/remove columns

gregwdata:
Ran into a bug in incremental models with `on_schema_change='append_new_columns'`. If the column names are invalid without quotes (e.g. have a space), the original `duckdb__alter_relation_add_remove_columns` macro inserting them raw would throw an error. 

* Updated the macro to use the dbt [Relation](https://docs.getdbt.com/reference/dbt-classes#creating-relations) method to respect the project's identifier quoting config.
* Added two tests:
  * First test confirms that incremental model adding a column with a space in the name **fails** if identifier quoting is `false`.
  * Second test confirms incremental add column **succeeds** if identifier quoting is `true`
  * Had to set up two new test classes from the `BaseIncrementalOnSchemaChangeSetup` class to be able to override project-level settings without stepping on (or redundantly running) the other tests in `BaseIncrementalOnSchemaChange`

### DIFF
--- a/dbt/include/duckdb/macros/columns.sql
+++ b/dbt/include/duckdb/macros/columns.sql
@@ -4,7 +4,7 @@
     {% for column in add_columns %}
       {% set sql -%}
          alter {{ relation.type }} {{ relation }} add column
-           {{ column.name }} {{ column.data_type }}
+           {{ api.Relation.create(identifier=column.name) }} {{ column.data_type }}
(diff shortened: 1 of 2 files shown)


jwills:
looks great @gregwdata -- will approve/merge as soon as tests pass!

jwills (approved):

PR #550, opened 2025-05-15
Add named secret parameter to attachment class for profiles.yml

dumkydewilde:
As per the [DuckDB docs](https://duckdb.org/docs/stable/extensions/postgres.html#managing-multiple-secrets), a secret can be given a name (which is already possible) and this secret can be referenced in the attach statement. This is convenient if you want to e.g. attach multiple postgres databases


jwills (approved):
LG-- thanks!

PR #553, opened 2025-05-19
Read .duckdbrc file on DuckDBT CLI startup

elephantmetropolis:
https://github.com/duckdb/dbt-duckdb/issues/551

I went with the simplest solution to replicate DuckDB CLI behavior. If `.duckdbrc` exists, we run its content, otherwise we skip it. It's minimal but I think it do the job.


jwills:
looks good @elephantmetropolis -- assuming you tested this out locally and it works as intended?

elephantmetropolis:
Sorry I forgot to precise, yes naturally I tested locally and it worked as expected !

PR #554, opened 2025-05-20
See if this is related to the mysterious doc test failures that popped up on master

jwills:
These doc test failures just started happening on nightly and I'm not quite sure why, so seeing if this is related.


matsonj (approved):
reviewed and approved

PR #556, opened 2025-05-27
Add support for arbitrary key-value pairs in Attachment ATTACH options

jwills:
## Summary
• Adds support for arbitrary key-value pairs in database attachment configuration via a new `options` dictionary
• Maintains backward compatibility with existing direct field configuration (`type`, `secret`, `read_only`)
• Provides conflict detection to prevent specifying the same option in both direct fields and options dict

## Key Changes
- Added `options` field to `Attachment` class for arbitrary ATTACH options
- Updated `to_sql()` method to handle both legacy fields and new options dict
- Added comprehensive test coverage for new functionality and conflict detection
- Updated README documentation with examples of new configuration format

## Use Case
This enhancement allows users to leverage new DuckDB attachment features without waiting for explicit support in dbt-duckdb. For example, users can now specify cache sizes, thread counts, and other advanced options that DuckDB supports in ATTACH statements.

## Test plan
- [x] Added unit tests for options dict functionality
- [x] Added tests for conflict detection between legacy fields and options
- [x] Verified backward compatibility with existing configurations
- [x] Updated documentation with clear examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,8 +168,8 @@ When fetching a secret for a path, the secret scopes are compared to the path, r
 
 #### Attaching Additional Databases
 
-DuckDB version `0.7.0` added support for [attaching additional databases](https://duckdb.org/docs/sql/statements/attach.html) to your dbt-duckdb run so that you can read
-and write from multiple databases. Additional databases may be configured using [dbt run hooks](https://docs.getdbt.com/docs/build/hooks-operations) or via the `attach` argument
+DuckDB supports [attaching additional databases](https://duckdb.org/docs/sql/statements/attach.html) to your dbt-duckdb run so that you can read
+and write from multiple databases. Additional databases may be configured via the `attach` argument
 in your profile that was added in dbt-duckdb `1.4.0`:
 
 ```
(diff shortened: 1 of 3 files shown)


wideltann:
This works as long as there's no slash in the attach or the data_path. Here's an example section of profiles.yml

```yaml
      attach:
        - path: ducklake:sqlite:storage/metadata.sqlite
          alias: dl
          options:
            DATA_PATH: storage/data_files
```
which produces the error: 
```
Runtime Error
  Parser Error: syntax error at or near "/"
```

Another issue arises if you set the database for the profile:
as in `database: dl`

```
21:33:00  Running with dbt=1.9.4
21:33:01  Encountered an error:
Runtime Error
  Credentials in profile "dds", target "default" invalid: Runtime Error
    Inconsistency detected between 'path' and 'database' fields in profile; the 'database' property must be set to 'memory' to match the 'path'
```

wideltann:
Claude suggests:
```python
<<<<<<< SEARCH
        elif path_db and data["database"] != path_db:
            if not data.get("remote"):
                raise DbtRuntimeError(
                    "Inconsistency detected between 'path' and 'database' fields in profile; "
                    f"the 'database' property must be set to '{path_db}' to match the 'path'"
                )
=======
        elif path_db and data["database"] != path_db:
            # Skip this validation when using attach configuration
            if not data.get("remote") and not data.get("attach"):
                raise DbtRuntimeError(
                    "Inconsistency detected between 'path' and 'database' fields in profile; "
                    f"the 'database' property must be set to '{path_db}' to match the 'path'"
                )
>>>>>>> REPLACE
```
which seems reasonable

matsonj (approved):
Looks good, once we have a better understanding of Duck Lake I assume we will have more work (i.e. attach the catalog)

PR #557, opened 2025-05-28
Update macros for dropping schema and relations when using ducklake

wideltann:
This PR implements special handling for ducklake relations in the duckdb__drop_relation and
duckdb__drop_schema macros.

Changes:

 • The duckdb__drop_relation macro now detects ducklake relations by checking if the
   relation's database matches an attachment alias with a ducklake path
 • The duckdb__drop_schema macro uses the same detection logic
 • For ducklake relations/schemas, we drop without the CASCADE option
 • For non-ducklake relations/schemas, we use CASCADE when dropping

Ducklake relations require special handling when being dropped. This implementation ensures
that ducklake relations are properly detected based on the attachment configuration in the
profile.

This might be too much of an edit for the adapters.sql but it does fix the cascade issue. I tested by creating these macros in my own dbt project and then testing the table materialization. I'm not totally sure when the schema drop gets called but I added the same logic there anyway.

### DIFF
--- a/dbt/include/duckdb/macros/adapters.sql
+++ b/dbt/include/duckdb/macros/adapters.sql
@@ -164,11 +164,7 @@ def materialize(df, con):
 
 {% macro duckdb__drop_relation(relation) -%}
   {% call statement('drop_relation', auto_begin=False) -%}
-    {% if relation.path is defined and relation.path.startswith('ducklake:') %}
-      {{ log("Dropping ducklake relation without cascade: " ~ relation, info=True) }}
       drop {{ relation.type }} if exists {{ relation }}
-    {% else %}
-      drop {{ relation.type }} if exists {{ relation }} cascade
     {% endif %}
   {%- endcall %}
 {% endmacro %}
--- a/dbt/include/duckdb/macros/adapters.sql
+++ b/dbt/include/duckdb/macros/adapters.sql
@@ -164,7 +164,7 @@ def materialize(df, con):
 
 {% macro duckdb__drop_relation(relation) -%}
   {% call statement('drop_relation', auto_begin=False) -%}
-      drop {{ relation.type }} if exists {{ relation }}
+      drop {{ relation.type }} jif exists {{ relation }}
     {% endif %}
   {%- endcall %}
 {% endmacro %}
--- a/dbt/include/duckdb/macros/adapters.sql
+++ b/dbt/include/duckdb/macros/adapters.sql
@@ -164,7 +164,11 @@ def materialize(df, con):
 
 {% macro duckdb__drop_relation(relation) -%}
   {% call statement('drop_relation', auto_begin=False) -%}
-      drop {{ relation.type }} jif exists {{ relation }}
+    {% if relation.path is defined and relation.path.startswith('ducklake:') %}
+      {{ log("Dropping ducklake relation without cascade: " ~ relation, info=True) }}
+      drop {{ relation.type }} if exists {{ relation }}
+    {% else %}
+      drop {{ relation.type }} if exists {{ relation }} cascade
     {% endif %}
   {%- endcall %}
 {% endmacro %}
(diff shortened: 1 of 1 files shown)


wideltann:
PR Title:

  refactor: move ducklake detection to @available adapter method with comprehensive tests

  PR Description:

  ## Summary

  Refactors ducklake database detection logic from Jinja macros to a proper `@available` adapter method, improving code organization, testability, and following dbt-duckdb
  architectural conventions.

  ## Changes

  ### Core Refactor
  - **Move logic to adapter class**: Added `is_ducklake()` method to `DuckDBAdapter` with `@available` decorator
  - **Remove code duplication**: Eliminated repeated ducklake detection logic between `drop_schema` and `drop_relation` macros
  - **Improve type safety**: Pass full `DuckDBRelation` object instead of parsing strings
  - **Follow conventions**: Use recommended `@available` pattern for complex adapter functionality

  ### Comprehensive Testing
  - **9 comprehensive test cases** covering all scenarios:
    - No attach configuration
    - Empty attach configuration
    - Valid ducklake attachments (`ducklake:sqlite:`, `ducklake:postgres:`)
    - Regular (non-ducklake) attachments
    - Mixed attachment configurations
    - Edge cases: None relation, missing database, missing alias, empty paths
  - **All tests pass** ✅

  ### Code Quality
  - **13 net lines removed** while improving maintainability
  - **Single source of truth** for ducklake detection logic
  - **Better error handling** with proper attribute checking
  - **Easier to extend** for future ducklake features

  ## Before/After

  **Before (Jinja macro approach):**
  ```sql
  {% if execute %}
    {% set profile = adapter.config.credentials %}
    {% if profile.attach is defined and profile.attach %}
      {% for attachment in profile.attach %}
        {% if attachment.alias == relation.database and attachment.path is defined and 'ducklake:' in attachment.path %}
          // ducklake logic
        {% endif %}
      {% endfor %}
    {% endif %}
  {% endif %}

  After (Adapter method approach):
  {% if adapter.is_ducklake(relation) %}
    // ducklake logic
  {% endif %}

  Testing

  python -m unittest tests.unit.test_duckdb_adapter.TestDuckDBAdapterIsDucklake -v
  # Ran 9 tests in 0.882s - OK

  Addresses

  Maintainer feedback to use @available adapter methods for complex logic instead of Jinja macros, improving testability and code organization.

dbt/include/duckdb/macros/adapters.sql
```diff
@@ -164,7 +168,11 @@ def materialize(df, con):
 
 {% macro duckdb__drop_relation(relation) -%}
   {% call statement('drop_relation', auto_begin=False) -%}
-    drop {{ relation.type }} if exists {{ relation }} cascade
+    {% if adapter.is_ducklake(relation) %}
```
jwills, line 171: yeah this is much cleaner

dbt/adapters/duckdb/impl.py
```diff
@@ -69,6 +69,24 @@ def debug_query(self):
     def is_motherduck(self):
         return self.config.credentials.is_motherduck
 
+    @available
+    def is_ducklake(self, relation: DuckDBRelation) -> bool:
+        """Check if a relation's database is backed by a ducklake attachment."""
+        if not relation or not relation.database:
+            return False
+        
+        attach_configs = getattr(self.config.credentials, 'attach', None)
```
jwills, line 78: So for this bit, I would rather have just a `set` on the Credentials class that contained the names of the databases that were ducklakes so we could just do a `relation.database is in self.config.credentials._ducklake_dbs` vs. looping over the attached databases each time. (My goal here is minimizing the overhead of this for dbt-duckdb projects that don't use ducklakes.) You'll want to put a block for this in the `__post_init__` function in `Credentials`.

jwills (commented):
Looking good-- one question + one thing I need fixed

dbt/adapters/duckdb/impl.py
wideltann: makes sense! we want that O(1) lookup

jwills (approved):
Looks great-- thank you so much @wideltann!

PR #571, opened 2025-06-09
Assume native comment support in DuckDB catalog queries

jwills:
## Summary
- Removes the custom `catalog_comment` adapter method from the DuckDB adapter
- Updates catalog SQL macros to directly use DuckDB's native `comment` field instead of the adapter method

## Test plan
- [ ] Run existing catalog-related tests to ensure comment retrieval still works
- [ ] Verify that table and column comments are properly displayed in dbt docs


jwills:
just a bit of cleanup here, no functionality change. We no longer build/test against 0.10.1 (we require a minimum of 1.0.0) so this condition can't happen anymore

PR #577, opened 2025-06-19
Fix bug where schema's database not specified

### LINKED ISSUE #576: Transaction error in `register_upstream_external_models` when using `database` target
When using an aliased database target, `register_upstream_external_models` will throw an error because it attempts to create a schema under the default database (e.g. `:memory:`) and then create a schema under the named database in the same transaction. This will fail because you can only modify a single database in one transaction. Instead, the create schema call should reference the target database.

Example:

profiles.yml:
```yaml
my_profile:
  outputs:
    ducklake_local:
      extensions:
      - ducklake
      type: duckdb
      attach:
        - path: ducklake:metadata.ducklake
          alias: ducklake
          options:
            DATA_PATH: "'data_files'"
      database: ducklake
```

dbt_project.yml:
```yaml
on-run-start:
  - "{{ register_upstream_external_models() }}"
```

Output:
```
Runtime Error
  TransactionContext Error: Attempting to write to database "ducklake" in a transaction that has already modified database ":memory:" - a single transaction can only write to a single attached database.
```

tehunter:
Closes #576

### DIFF
--- a/dbt/include/duckdb/macros/utils/upstream.sql
+++ b/dbt/include/duckdb/macros/utils/upstream.sql
@@ -22,7 +22,7 @@
         {%- set upstream_location = adapter.external_read_location(location, rendered_options) -%}
         {% if upstream_rel.schema not in upstream_schemas %}
           {% call statement('main', language='sql') -%}
-            create schema if not exists {{ upstream_rel.schema }}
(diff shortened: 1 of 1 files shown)


jwills (approved):
Thank you Thomas!

PR #579, opened 2025-06-20
Fix error swallowing that hides DuckDB exception details

jwills:
## Summary
- Fixed RuntimeError exceptions being swallowed in exception_handler, preventing users from seeing actual DuckDB error details
- Preserved original exception context in DuckDBCursorWrapper by maintaining exception chain with 'from e'
- Added debug logging for commit failures instead of silent swallowing to improve debugging visibility

## Problem
Users were experiencing generic "TransactionContext Error: Current transaction is aborted (please ROLLBACK)" messages instead of seeing the actual underlying DuckDB errors that caused the transaction to fail. This made debugging difficult, especially for MotherDuck users.

## Root Cause
Three locations in the codebase were swallowing or losing error context:
1. `connections.py` exception_handler was catching RuntimeError but only logging without re-raising
2. `local.py` DuckDBCursorWrapper was converting RuntimeError to string, losing stack trace
3. `impl.py` commit_if_has_connection was silently ignoring DbtInternalError

## Solution
- Preserve full exception chains using `from e` syntax to maintain original error context
- Re-raise RuntimeError with original details instead of swallowing
- Add debug logging for commit failures while maintaining backward compatibility

## Test plan
- [ ] Test with intentionally failing SQL to verify original error messages are preserved
- [ ] Verify MotherDuck constraint violations show detailed error messages
- [ ] Confirm transaction errors display underlying cause instead of generic messages
- [ ] Run existing test suite to ensure no regressions

### DIFF
--- a/dbt/adapters/duckdb/impl.py
+++ b/dbt/adapters/duckdb/impl.py
@@ -194,7 +194,7 @@ class DuckDBAdapter(SQLAdapter):
             self.connections.commit_if_has_connection()
         except DbtInternalError as e:
             # Log commit errors instead of silently swallowing them to aid debugging
-            logger.debug(f"Commit failed with DbtInternalError: {e}")
+            logger.exception(f"Commit failed with DbtInternalError: {e}")
             # Still pass to maintain backward compatibility, but now with visibility
             pass
 


jwills:
@rchui mi tokens es su tokens

hrl20 (approved):

PR #591, opened 2025-07-08
Fix incremental Python models on MotherDuck

hiddestokvis:
I ran into issues with incremental Python models. The temp table created in dbt_temp is created after the query is executed leading into a table not found error. As far as I can see this seems a small omission in the create_table_as function when the language is set to Python, although I am uncertain if this is done for a specific reason. Nothing seems to be mentioned here: https://github.com/duckdb/dbt-duckdb/pull/326.


jwills:
@hiddestokvis looks like an oversight to me-- thanks for the fix here! cc @hrl20

Issue #587, opened 2025-06-30
New exception log statement triggers dbt-dagster failures


paultiq:
https://github.com/duckdb/dbt-duckdb/commit/f95af269001da7eee4a43e3c51462a36b998121e https://github.com/duckdb/dbt-duckdb/pull/579 added a `logger.exception` call that emits an exception fairly frequently, at least for me.

https://github.com/duckdb/dbt-duckdb/blame/1f7e6abc82442191b83ba49a0bee722b0d46c99c/dbt/adapters/duckdb/impl.py#L193C13-L193C16

dbt-dagster treats this emitted exception as job failure, even with a normal return code. This is because dbt-dagster treats any error severity message as a failure. 

Since this is a message is probably benign, downgrading it to a warning seems appropriate. 

I've fixed locally with: `logger.warning(f"Commit failed with DbtInternalError: {e}")`

```
failed with exit code `0`. Check the stdout in the Dagster compute logs for the full information about the error, or view the dbt debug log: /git/<myrepo>/<myproject>/target/my_dbt_assets3-3bd74a9-7f5c095/dbt.log.

Errors parsed from dbt logs:

DuckDB adapter: Commit failed with DbtInternalError: Internal Error
  Tried to commit transaction on connection "master", but it does not have one open!

Stack Trace:
  File "/git/<myrepo>/.venv/Lib/site-packages/dagster/_core/execution/plan/execute_plan.py", line 246, in dagster_event_sequence_for_step
    yield from check.generator(step_events)
  File "/git/<myrepo>/.venv/Lib/site-packages/dagster/_core/execution/plan/execute_step.py", line 504, in core_dagster_event_sequence_for_step
    for user_event in _step_output_error_checked_user_event_sequence(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/git/<myrepo>/.venv/Lib/site-packages/dagster/_core/execution/plan/execute_step.py", line 185, in _step_output_error_checked_user_event_sequence
    for user_event in user_event_sequence:
                      ^^^^^^^^^^^^^^^^^^^
  File "/git/<myrepo>/.venv/Lib/site-packages/dagster/_core/execution/plan/execute_step.py", line 88, in _process_asset_results_to_events
    for user_event in user_event_sequence:
```

[...]

PR #612, opened 2025-08-21
Fix orchestrator compatibility: change logger.exception to logger.warning

jwills:
## Summary
• Replace logger.exception with logger.warning + traceback formatting to preserve stack trace info
• Fixes compatibility issue with orchestrators like Dagster that treat logger.exception calls as fatal exceptions

## Test plan
• All existing unit tests continue to pass (92 tests)
• Pre-commit hooks pass (mypy, ruff, etc.)
• Stack trace information is preserved in the warning log output

### DIFF
--- a/dbt/adapters/duckdb/impl.py
+++ b/dbt/adapters/duckdb/impl.py
@@ -1,4 +1,5 @@
 import os
+import traceback
 from collections import defaultdict
 from typing import Any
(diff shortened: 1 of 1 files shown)


jwills:
Fixes all of the pain and suffering from #587 at the cost of possibly introducing new sources of pain and suffering

PR #621, opened 2025-09-03
Display traceback when Python model raises exception

thomas-daniels:
When a Python model errors, only the error message would be displayed, not the traceback. That makes it more difficult to debug more complicated Python models. This commit makes sure that the entire traceback gets displayed.

### DIFF
--- a/dbt/adapters/duckdb/environments/__init__.py
+++ b/dbt/adapters/duckdb/environments/__init__.py
@@ -289,7 +289,7 @@ class Environment(abc.ABC):
                 module.materialize(df, cur)
         except Exception as err:
             raise DbtRuntimeError(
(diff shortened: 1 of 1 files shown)


thomas-daniels:
Thank you for your response! The problem did not seem to be the double quotes (this also worked fine when I tried it out), but the fact that the line was overly long. Should be fixed now, linting passes locally.

jwills:
@thomas-daniels nah, still seems like something is off-- maybe a python version thing here? It looks like a syntax diff in 3.12 vs. 3.10 etc. etc. and I try to do things in a way that we maintain compat with older versions of python: https://docs.getdbt.com/faqs/Core/install-python-compatibility

thomas-daniels:
oh I see, that makes sense, I was indeed testing only with Python 3.12. Now it should be good.

PR #622, opened 2025-09-03
Replace "type" key with "managed_ducklake_dbs" key for configuring MotherDuck profiles.

matsonj:
## Summary
Introduce explicit, granular detection for MotherDuck-managed DuckLake catalogs via a new profile option, avoid invalid ATTACH behavior on MD, and skip constraints for DuckLake relations.

## Why
- Previous implementation ( #620) caused issue with attachment-level `type: ducklake` on `md:` causes DuckDB to attempt DuckLake initialization, which fails on MotherDuck and triggers PRIMARY KEY/UNIQUE errors.
- Aliasing and attach_mode constraints in MD make per-attachment detection brittle.
- We need a reliable way to identify managed DuckLake catalogs to:
 - Disable CASCADE on drops
 - Skip constraints rendering (PK/UNIQUE) during CTAS
 
 ## Profile Examples
 
#### MotherDuck-managed DuckLake

Note the new key: `ducklake_managed_dbs`.

```yml
dbt_ducklake:
  target: ducklake
  outputs:
    ducklake:
      type: duckdb
      path: 'md:dbt_ducklake'          
      threads: 4
      ducklake_managed_dbs:
        - dbt_ducklake             
```

#### Local DuckLake

Existing Ducklake detection from #557 works without issue here.

```yml
outputs:
  local:
    type: duckdb
    path: ':memory:'
    attach:
      - alias: lk
        path: 'ducklake:sqlite:storage/metadata.sqlite'
        options:
          data_path: 's3://your-bucket/prefix'
```
## Testing
Unit and functional tests updated to validate `ducklake_managed_dbs`-based detection.


matsonj:
closing in favor of #625

PR #624, opened 2025-09-05
duckdb: fix dateadd interval syntax for newer DuckDB versions

jwills:
This updates the DuckDB dateadd macro to use the standard `INTERVAL <n> <unit>` form rather than `interval (<n>) <unit>`. In upcoming DuckDB releases the parenthesized literal form can raise:\n\n> Not implemented Error: Unimplemented expression class\n\nThis change keeps date_spine/dbt_utils flows working on nightly and remains compatible with current stable releases.


jwills:
@hrl20 I verified that this fixes the failures we were seeing in the nightly builds on a linux box by running `tox -e nightly -- tests/functional/adapter/utils/test_date_spine.py -q`

hrl20 (approved):

PR #638, opened 2025-09-23
Break out create temp schema into its own transaction

hrl20:
addresses https://github.com/duckdb/dbt-duckdb/issues/636

The default incremental Python model (intermediate result) materialization path will [use a new cursor to create a temp table](https://github.com/duckdb/dbt-duckdb/blob/ae5c63f3658f6c296030a07527f61473367e4d0f/dbt/adapters/duckdb/environments/__init__.py#L282-L289). 

This does not work when connected to MotherDuck and with transaction enabled (the setting is `disable_transaction: false`- which is the default), because the incremental model transaction currently wraps all of (`create temp schema`, `delete from target table`, `insert into target table`, `drop temp table`) inside one transaction, and the result of the materialization query, which happens between `create temp schema` and `delete from target table`, will not be visible to the transaction. 

The reason this works for local DuckDB is that local DuckDB uses a real temp table, which does not require a temp schema to be created, so the transaction starts after the materialiaztion query. 

There's no strong reason to include the create temp schema query in the transaction, so we break it out of the main transaction, such that the main transaction starts after the materialization query for the MotherDuck code path.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -81,7 +81,7 @@ def dbt_profile_target(profile_type, bv_server_process, tmpdir_factory):
             profile["token"] = os.environ.get(TEST_MOTHERDUCK_TOKEN)
         else:
             profile["token"] = os.environ.get(MOTHERDUCK_TOKEN, os.environ.get(MOTHERDUCK_TOKEN.lower()))
-        profile["disable_transactions"] = False
+        profile["disable_transactions"] = True
         profile["path"] = "md:test"
     elif profile_type in ["memory", "nightly"]:
         pass  # use the default path-less profile


jwills:
@hrl20 the MD test still fails on CI tho? Is it unrelated, or something?

hrl20:
> > @hrl20 the MD test still fails on CI tho? Is it unrelated, or something?
> 
> :/ seems to be failing consistently but not apparent why. looking.

these tests seems to be failing on main too when transaction is enabled in the test profile. i'm going to keep transaction disabled here and fix separately.

jwills (approved):

PR #677, opened 2025-12-16
Log duckdb Error Message

Tseegi-N:
A very small PR on the exception handler as people have reported not being able to see DuckDB error messages. (MotherDuck Support)

Details:
The `exception_handler` logs errors for `RuntimeError` but not for other exceptions. I'm adding error logging there to be consistent with the `RuntimeError` handler.

Some reasons of the top of my head on why error messages need to be preserved with full context:
1. The error message is hard to find (buried in logs)
2. It's separated from the SQL that caused it
3. You have to manually correlate "Error running SQL: COMMIT" with the error message that appears much later in the logs


jwills:
Thanks @Tseegi-N !

Tseegi-N:
Thank you, @jwills! Will the changes be effective in the next release? If so, do you have an estimated timeline for that? Just wanted to know for user communication.

PR #644, opened 2025-10-08
feat(duckdb): add MERGE and microbatch support for incremental strategies

gfrmin:
Added version-aware support for MERGE statements in DuckDB adapter for versions 1.4.0 and above. Introduced `_supports_merge` property to check and cache version compatibility, enabling "merge" and "microbatch" incremental strategies alongside existing "append" and "delete+insert". Updated SQL macros to implement microbatch and merge logic for efficient batch processing and upserts.

PR #681, opened 2025-12-23
Microbatch alternative

dumkydewilde:
There's been a good effort from @gfrmin to implement microbatching in dbt-duckdb in #644. However, I wanted to show an approach that I believe is more reliable and future proof. 

This PR adds a microbatching specific incremental strategy that is similar to delete+insert, but slightly different. It uses the batch context (`model.get('batch')`) to set the batch start and end as the first predicates. This allows us to filter:
- incoming data from the source
- existing data when deleting
- inserts

Performance is best when the `event_time` column matches the granularity of the batch_size. 

Benchmark results (from: https://github.com/dumkydewilde/dbt-duckdb-clickbench )

|             model_name              |          target_name            | model_runtime_s |  
|-------------------------------------|--------------------------------|-------------|
| table                               | incremental_multi_thread_8_8GB |       26.692926 | 
| table                               | incremental_multi_thread_3_8GB |         33.4093 | 
| microbatch_event_date_col           | incremental_multi_thread_3_8GB |        60.26188 | 
| microbatch_ukey                     | incremental_multi_thread_8_8GB |        75.95332 | 
| microbatch_default                  | incremental_multi_thread_8_8GB |        76.62632 | 
| incremental__del_ins_date_partition | incremental_multi_thread_8_8GB |       77.524574 | 
| microbatch_ukey_date_partition      | incremental_multi_thread_3_8GB |       79.064766 | 
| microbatch_ukey                     | incremental_multi_thread_3_8GB |        79.11033 | 
| incremental__del_ins_ukey_date      | incremental_multi_thread_8_8GB |        80.00094 | 
| microbatch_date_partition           | incremental_multi_thread_8_8GB |        80.69873 | 
| microbatch_ukey_date_partition      | incremental_multi_thread_8_8GB |        81.39617 | 
| microbatch_default                  | incremental_multi_thread_3_8GB |        85.59948 | 
| incremental__del_ins_date_partition | incremental_multi_thread_3_8GB |       87.599686 | 
| microbatch_date_partition           | incremental_multi_thread_3_8GB |        90.62669 | 
| incremental__del_ins_ukey_date      | incremental_multi_thread_3_8GB |      118.220695 | 
| incremental__del_ins_ukey           | incremental_multi_thread_8_8GB |       120.06944 | 
| incremental__del_ins_ukey           | incremental_multi_thread_3_8GB |       124.26179 | 
| table                               | incremental_single_thread_8GB  |       127.54384 | 
| microbatch_default                  | incremental_single_thread_8GB  |       182.72571 | 
| microbatch_ukey_date_partition      | incremental_single_thread_8GB  |       182.82524 | 
| microbatch_ukey                     | incremental_single_thread_8GB  |         185.796 | 
| microbatch_date_partition           | incremental_single_thread_8GB  |       205.63583 | 
| incremental__del_ins_date_partition | incremental_single_thread_8GB  |       245.33249 | 
| incremental__del_ins_ukey           | incremental_single_thread_8GB  |        324.3013 | 
| incremental__del_ins_ukey_date      | incremental_single_thread_8GB  |       327.11945 |

### DIFF
--- a/dbt/include/duckdb/macros/materializations/incremental_strategy/microbatch.sql
+++ b/dbt/include/duckdb/macros/materializations/incremental_strategy/microbatch.sql
@@ -49,7 +49,8 @@
 
         insert into {{ target }} ({{ dest_cols_csv }})
         select {{ dest_cols_csv }}
-        from {{ source }};
+        from {{ source }}
+        where {{ batch_predicate }};
     {%- endset -%}
 
     {{ return(build_sql) }}
--- a/tests/functional/adapter/test_incremental_microbatch.py
+++ b/tests/functional/adapter/test_incremental_microbatch.py
@@ -41,13 +41,13 @@ select 1 as id, '2025-01-01 00:00:00'::timestamp as event_time
 
 # Scenario models
 models__microbatch_exec_input = """
-{{ config(materialized='table', event_time='event_time') }}
+{{ config(materialized='table') }}
 
-select 1 as id, '2025-01-01'::date as event_time
+select 1 as id, '2025-01-01'::date as event_ts
 union all
-select 2 as id, '2025-01-02'::date as event_time
+select 2 as id, '2025-01-02'::date as event_ts
 union all
-select 3 as id, '2025-01-03'::date as event_time
+select 3 as id, '2025-01-03'::date as event_ts
 """
 
 models__microbatch_exec = """
@@ -59,7 +59,7 @@ models__microbatch_exec = """
     begin=modules.datetime.datetime(2025, 1, 1, 0, 0, 0)
 ) }}
 
-select id, event_time from {{ ref('microbatch_exec_input') }}
+select id, event_ts as event_time from {{ ref('microbatch_exec_input') }}
 """
 
 models__microbatch_event_date_input = """
@@ -331,7 +331,7 @@ class TestMicrobatchScenarios:
 
         source = relation_from_name(project.adapter, "microbatch_exec_input")
         project.run_sql(
-            f"insert into {source} (id, event_time) values "
+            f"insert into {source} (id, event_ts) values "
             "(4, '2025-01-02 12:00:00'::timestamp), "
             "(5, '2025-01-03 12:00:00'::timestamp)"
         )


dbt/include/duckdb/macros/materializations/incremental_strategy/microbatch.sql
```diff
@@ -0,0 +1,50 @@
...
+    {# Extract remaining arguments #}
+    {%- set target = arg_dict["target_relation"] -%}
+    {%- set source = arg_dict["temp_relation"] -%}
+    {%- set dest_columns = arg_dict["dest_columns"] -%}
+    {%- set incremental_predicates = validate_incremental_predicates(arg_dict.get("incremental_predicates")) -%}
+    {%- set dest_cols_csv = get_quoted_csv(dest_columns | map(attribute="name")) -%}
+
+    {# Build the batch time filter predicate #}
+    {%- set batch_predicate -%}
+        {{ event_time }} >= cast('{{ batch_start }}' as timestamp)
```
cobac: instead of casting to ts, compare types of `event_time` and `batch_xxx` and emit a warning if they would force a full table scan (?)

tests/functional/adapter/test_incremental.py
```diff
@@ -143,6 +152,48 @@ def test__handle_identifier_quoting_config_false(self, project):
         assert status == RunStatus.Success
 
 
+@pytest.mark.skipif(not MICROBATCH_AVAILABLE, reason="Microbatch tests require dbt-core >= 1.9")
```
hrl20: would be good to have tests verifying idempotency under various configurations (with/wo unique key)- running microbatch in incremental mode several times over the same time range and verifying the results don't change.

dbt/include/duckdb/macros/materializations/incremental_strategy/microbatch.sql
```diff
@@ -0,0 +1,50 @@
...
+        and {{ event_time }} < cast('{{ batch_end }}' as timestamp)
```
hrl20: I'm concerned that always casting to `timestamp` won't work when user's event time column is `timestamptz`. some ideas:
1) perhaps not casting, and just using the value of the batch_start/_end literal, could work (by forcing duckdb to infer and casting the timestamp to the right type). but it may also give rise to subtle & surprising behaviors with timezones (especially given this is hidden from the user). 
2) make this an option, so it's explicitly controlled by the user. 
3) figure out the type of event_time dynamically?

tests/functional/adapter/test_incremental.py
hrl20: bonus: test against ducklake

dbt/include/duckdb/macros/materializations/incremental_strategy/microbatch.sql
```diff
@@ -0,0 +1,50 @@
+{% macro duckdb__get_incremental_microbatch_sql(arg_dict) -%}
+    {# Extract and validate required config #}
```
hrl20: does this implementation make use of user-supplied unique_key? if not, which is fine by me, it would be good to explicitly disable that so users don't think unique_key has any effect.

dbt/include/duckdb/macros/materializations/incremental_strategy/microbatch.sql
hrl20: good call out on the UTC conversion, if that's the case then i think using the literal should just works (a test case to check the edge cases would be nice)
the other concern is performance one. It looks like the cast to timestamp on the RHS will force a cast on the event_time column. 

Here's the difference on the `microbatch_event_date` model
with ::timestamp
```
┌───────────────────────────┐
│   HYBRID_STATS_COLLECTOR  │
│    ────────────────────   │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│       HYBRID_RUNNER       │
│    ────────────────────   │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│      DOWNLOAD_BRIDGE      │
│    ────────────────────   │
│           ~1 row          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│           DELETE          │
│    ────────────────────   │
│     Expressions: rowid    │
│                           │
│           ~1 row          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│  MD_REMOTE_CATALOG_TABLE  │
│    ────────────────────   │
│          Filters:         │
│    (CAST(event_date AS    │
│  TIMESTAMP) BETWEEN '2025 │
│     -01-03 00:00:00':     │
│ :TIMESTAMP AND '2025-01-04│
│    00:00:00'::TIMESTAMP)  │
│                           │
│           Table:          │
│md:test.test176789458493971│
│00151_test_incremental_micr│
│obatch.microbatch_event_dat│
│             e             │
│                           │
│           ~1 row          │
└───────────────────────────┘
```
vs without `::timestamp`
```
┌───────────────────────────┐
│   HYBRID_STATS_COLLECTOR  │
│    ────────────────────   │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│       HYBRID_RUNNER       │
│    ────────────────────   │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│      DOWNLOAD_BRIDGE      │
│    ────────────────────   │
│           ~1 row          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│           INSERT          │
│    ────────────────────   │
│           ~1 row          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         PROJECTION        │
│    ────────────────────   │
│        Expressions:       │
│             id            │
│         event_date        │
│                           │
│           ~1 row          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│  MD_REMOTE_CATALOG_TABLE  │
│    ────────────────────   │
│          Filters:         │
│ event_date>='2025-01-03': │
│ :DATE AND event_date<'2025│
│       -01-04'::DATE       │
│                           │
│           Table:          │
│      md:test.dbt_temp     │
│.microbatch_event_date__309│
│         3321e8f34         │
│                           │
│           ~1 row          │
└───────────────────────────┘
```

dbt/include/duckdb/macros/materializations/incremental_strategy/microbatch.sql
```diff
@@ -0,0 +1,50 @@
...
+
+    {# Generate delete + insert SQL #}
+    {%- set build_sql -%}
+        delete from {{ target }}
+        where {{ where_clause }};
+
+        insert into {{ target }} ({{ dest_cols_csv }})
+        select {{ dest_cols_csv }}
+        from {{ source }}
+        where {{ where_clause }}
```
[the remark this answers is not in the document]
dumkydewilde: This was just indentation? I don't think the where statement should be indented more than the `from` right?

dbt/include/duckdb/macros/materializations/incremental_strategy/microbatch.sql
dumkydewilde: @hrl20 yeah, that was my finding as well. Casting can cause the query planner not to be able to filter row groups. I removed the casting altogether so it just uses the direct output of `batch_start` and `batch_end`. In the end the user can still make errors with this, so probably good to document it well.

hrl20 (approved):
thanks!!

PR #684, opened 2026-01-06
fix: rollback transaction in exception handler

hfrifkin:
The exception handler logged "Rolling back transaction" but never executed the rollback, leaving transactions open after errors. This can cause open idle transactions.


Donnype (approved):
As far as I can tell there is no rollback implemented anywhere else, and I suppose `rollback_if_open` means we don't have to check the `self.disable_transactions` setting, so this makes sense to me! 👍

PR #717, opened 2026-04-10
Fix single-threaded COPY in external materialization

### LINKED ISSUE #586: External materialization with PER_THREAD_OUTPUT performs single-threaded write
I recently noticed that using `external materialization` with `PER_THREAD_OUTPUT` resulted in the same write performance as without `PER_THREAD_OUTPUT`.

I investigated this and it seems that having the `COPY ...` operation in the same transaction block as the `CREATE TABLE ...` operation prevents DuckDB from using all available threads. In the case of large files being written to S3, this leads to very poor write performance (we have 40 threads available and plenty of bandwidth, yet only one thread is being used). I opened an [issue](https://github.com/duckdb/duckdb/issues/18074) on DuckDB to know if that behavior is "by design" or a bug. I tend to think it is "by design" even though I'm not sure I fully understand the implications of DuckDB's optimistic concurrency control.

If this is not a bug, I wanted to know if it was possible to change the behavior of `external materialization`. The idea would be to edit `external.sql` to end the first transaction before `write_to_file()` and open a new transaction (or not) after `write_to_file()` for the view creation and cleanup.

This would represent a shift from the current behavior, but it would be a trade-off between transactional atomicity and performance. As it stands today, external materialization cannot realistically be used for large data as the write overhead is too significant. We are currently using a workaround with macros and post-hooks to make it work, but it's not optimal.

I would greatly appreciate your input on this one @tomsej, as this part has not changed much since you started working on it. Also my proposed solution is probably not optimal and you may have a better idea to solve that issue.

jwills:
## Summary

- Commit the open transaction before `write_to_file` in the external materialization so that DuckDB can use multi-threaded writes for `COPY` operations
- Set `auto_begin=False` on the `write_to_file` statement to prevent auto-starting a new transaction before the COPY
- Add `per_thread_output` test case to the external materialization test suite

DuckDB forces single-threaded `COPY` when inside a transaction, making `PER_THREAD_OUTPUT` and `partition_by` writes ineffective for large data. This trades strict transactional atomicity around the COPY for usable write performance — the preexisting relation cleanup at the top of the materialization handles the partial-failure case.

Fixes #586

## Test plan

- [x] Existing external materialization tests pass with the transaction boundary change
- [x] New `per_thread_output` model exercises the option end-to-end
- [ ] Manual verification with large dataset + `PER_THREAD_OUTPUT` to confirm multi-threaded writes

### DIFF
--- a/dbt/include/duckdb/macros/adapters.sql
+++ b/dbt/include/duckdb/macros/adapters.sql
@@ -284,7 +284,7 @@ def materialize(df, con):
 {% endmacro %}
 
 {% macro write_to_file(relation, location, options) -%}
-  {% call statement('write_to_file') -%}
+  {% call statement('write_to_file', auto_begin=False) -%}
     copy {{ relation }} to '{{ location }}' ({{ options }})
   {%- endcall %}
 {% endmacro %}
--- a/dbt/include/duckdb/macros/materializations/external.sql
+++ b/dbt/include/duckdb/macros/materializations/external.sql
@@ -76,7 +76,15 @@
     {% endif %}
   {%- endcall %}
 
-  -- write a temp relation into file
+  -- Commit the transaction before COPY so DuckDB can use multi-threaded writes.
+  -- Without this, COPY inside a transaction is forced to use a single thread,
+  -- making PER_THREAD_OUTPUT and partition_by writes ineffective for large data.
+  -- See: https://github.com/duckdb/duckdb/issues/18074
+  {% if not adapter.disable_transactions() %}
+    {{ adapter.commit() }}
+  {% endif %}
+
+  -- write a temp relation into file (runs outside transaction for multi-threaded I/O)
   {{ write_to_file(temp_relation, location, write_options) }}
 
 -- create a view on top of the location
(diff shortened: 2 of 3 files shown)


hrl20 (approved):

steffenmllr:
Hey @jwills - I've tested this with `dbt-duckdb @ git+https://github.com/duckdb/dbt-duckdb.git@d14c227` and for me this still only used one core on a 48 core machine:

```
{{ config(
    materialized='external',
    location=/mnt/path/data,
    format='parquet',
    options={
        'compression': 'zstd',
        'per_thread_output': true,
        'overwrite_or_ignore': true
    }
) }}

SELECT * FROM {{ ref('mytable') }}
```

jwills:
hey @steffenmllr I can't repro that-- can you give me some more context? DuckDB version, make sure you haven't inadvertently set threads to 1 in DuckDB or are running inside of a cgroup that only has a single core, stuff like that?

jwills:
like I'm strongly starting to suspect that something else is going on here

steffenmllr:
yeah I think there is something else going on maybe using `env_var` in `profiles.yml` does not work as expected and sets this to `1` somehow - I'll investigate. Thanks for your work

PR #735, opened 2026-04-30
Allowing function expressions in partition_by

nicku33:
Previously, partition_by were assumed to be columns, which would always be escaped.
However `month(ts)`, if escaped would not be interpreted by DuckDB as an expression.
This logic for escaping now checks for a parentheses pair `..(..)` after a valid function name `[A-Za-z_]*` and does not escape in that case. To compensate for the risk of not escaping there are checks on SQL injection. People can still escape themselves if they need to with `month("ts")`

### DIFF
--- a/tests/unit/test_partition_render.py
+++ b/tests/unit/test_partition_render.py
@@ -1,0 +1,45 @@
+import pytest
+from dbt_common.exceptions import DbtRuntimeError
+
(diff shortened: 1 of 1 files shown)


nicku33:
Opened prematurely. Will test locally then reopen later.

PR #739, opened 2026-05-11
Skip expand_column_types on DuckDB to avoid slow DuckLake cold-start

### LINKED ISSUE #738: expand_target_column_types causes 60-90s cold-start on incremental models when DuckLake has a bloated Postgres Catalog (> 150k snapshots)
### Setup

dbt-duckdb 1.10.1, dbt 1.11.8, DuckDB 1.5.2
DuckLake attached via ducklake:postgres:dbname=..., data on S3
~4 schemas in the catalog, modest table count
One incremental model, merge strategy, on_schema_change='ignore', source is hive-partitioned parquet on S3

### Offending model xyz_bronze.sql
```sql
{# Schema changes are intentionally ignored — source parquet schema is stable.
   A full-refresh against a new hive location is the migration path if it ever changes. #}
{{ config(
    materialized='incremental',
    incremental_strategy='merge',
    unique_key='messageID',
    on_schema_change='ignore',
    tags=['xyz'],
) }}

WITH src AS (
    SELECT *
    FROM {{ source('xyz', 'xyz_raw') }}
    {% if is_incremental() %}
    WHERE modified >= (SELECT max(modified) FROM {{ this }}) - INTERVAL '1 day'
    {% endif %}
),

deduped AS (
    SELECT *,
        ROW_NUMBER() OVER (
            PARTITION BY messageID
            ORDER BY modified DESC, version DESC
        ) AS _rn
    FROM src
)

SELECT * EXCLUDE (_rn)
FROM deduped
WHERE _rn = 1
```

From source.yml

```yml
tables:
      - name: xyz_raw
        description: "Raw XYZ parquet files written by xyz ingest. Hive-partitioned by country."
        meta:
          external_location: "read_parquet('s3://project-data/xyz/country=*/*.parquet', hive_partitioning=true, union_by_name=true)"
          partition_keys: ["country"]
```

### Symptom
Model takes ~90s. Of that, ~78s is one query — the get_columns_in_relation lookup against the temp table created by the incremental:

```sql
select column_name, data_type, character_maximum_length, numeric_precision, numeric_scale
from system.information_schema.columns
where table_name = 'mymodel__dbt_tmp20260508122044098106'
order by ordinal_position
```

No table_schema / table_catalog filter because duckdb__make_temp_relation sets schema/database to none. With on_schema_change='ignore' set, the call still happens because expand_target_column_types runs unconditionally in the incremental materialization (line 67 of incremental.sql).

I run a py-spy flame graph and got:

```markdown
expand_target_column_types  (7,961 samples)
  └─ expand_column_types  (7,953)
     └─ get_columns_in_relation  (7,953)
        └─ execute_macro
           └─ ... jinja ...
              └─ cursor.execute(sql)  in environments/local.py:23  (7,953)
                 └─ ◀ samples drop off — DuckDB native code below
```

### What I tried
Running the exact same SQL in the duckdb CLI against the same DuckLake setup (after creating the temp table the same way and wrapping in BEGIN/ROLLBACK): 1.5–3s, not 78s. So the query can be fast — something in dbt-duckdb's runtime state pushes it ~25-50x slower.

jwills:
## Summary

Fixes #738. Overrides `DuckDBAdapter.expand_column_types` as a no-op so that incremental materializations no longer issue two `get_columns_in_relation` lookups whose results DuckDB cannot act on.

- DuckDB has no `VARCHAR(N)` length constraint, so `DuckDBColumn.can_expand_to` is always `False` (`string_size()` returns 256 for every string column). The base `SQLAdapter.expand_column_types` therefore never reaches its `alter_column_type` branch — only the column lookups remain, and they are pure overhead.
- On an incremental model against an attached DuckLake catalog backed by a bloated Postgres metastore (~150K+ snapshots in the reporter's case), the unfiltered `system.information_schema.columns where table_name = '<dbt_tmp...>'` lookup triggers a one-time DuckLake metastore load that can take 60–90s. After the cache is hot, the same query is ~0.08s.
- Diagnosis and proposed fix by @chaegordon in the issue.

## Test plan

- [x] New unit test `test_expand_column_types_is_noop` patches `get_columns_in_relation` and `alter_column_type` on the adapter and asserts neither is called.
- [x] Full unit suite passes (95 tests).
- [x] `tests/functional/adapter/test_incremental.py` passes (26 tests) on DuckDB 1.5.2.
- [x] `pre-commit run --all-files` clean.

### DIFF
--- a/dbt/adapters/duckdb/impl.py
+++ b/dbt/adapters/duckdb/impl.py
@@ -131,6 +131,14 @@ class DuckDBAdapter(SQLAdapter):
     def debug_query(self):
         self.execute("select 1 as id")
 
+    def expand_column_types(self, goal, current):
+        # DuckDB has no VARCHAR(N) length constraint, so DuckDBColumn.can_expand_to is
+        # always False (string_size() returns 256 for every string column). The base
+        # implementation's get_columns_in_relation lookups are pure overhead -- and can
+        # be extremely slow against attached catalogs (e.g. DuckLake on Postgres) where
+        # an unfiltered information_schema.columns scan triggers a full metastore load.
+        return
+
     @available
     def parse_index(self, raw_index: Any) -> Optional[DuckDBIndexConfig]:
         return DuckDBIndexConfig.parse(raw_index)
--- a/tests/unit/test_duckdb_adapter.py
+++ b/tests/unit/test_duckdb_adapter.py
@@ -65,6 +65,17 @@ class TestDuckDBAdapter(unittest.TestCase):
         self.adapter.connections.thread_connections[key] = mock_connection("main")
         self.assertEqual(len(list(self.adapter.cancel_open_connections())), 0)
 
+    def test_expand_column_types_is_noop(self):
+        # DuckDB has no VARCHAR(N), so the base SQLAdapter.expand_column_types is dead
+        # work -- and its unfiltered information_schema.columns lookup can be very slow
+        # against attached catalogs (e.g. DuckLake). Verify the override skips it.
+        with mock.patch.object(
+            self.adapter, "get_columns_in_relation"
+        ) as get_cols, mock.patch.object(self.adapter, "alter_column_type") as alter:
+            self.assertIsNone(self.adapter.expand_column_types(mock.MagicMock(), mock.MagicMock()))
+            get_cols.assert_not_called()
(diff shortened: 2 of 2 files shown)


hrl20 (approved):
nice

PR #749, opened 2026-05-22
Fix duckdb__create_schema for non-duckdb attached databases (closes #696)

### LINKED ISSUE #696: duckdb__create_schema fails for MySQL attached databases — missing type check like SQLite
## Description

When using `dbt-duckdb` with a MySQL database attached via DuckDB's `mysql` extension, `dbt run` fails at schema creation time with:

```
IO Error: Failed to execute MySQL query "CREATE SCHEMA `mydb`": Can't create database 'mydb'; database exists
```

## Root Cause

The [`duckdb__create_schema`](https://github.com/duckdb/dbt-duckdb/blob/main/dbt/include/duckdb/macros/adapters.sql) macro checks for `sqlite` type databases and skips schema creation, but does **not** handle `mysql` type databases:

```sql
{% set sql %}
    select type from duckdb_databases()
    where lower(database_name)='{{ relation.database | lower }}'
    and type='sqlite'    -- ⬅️ only checks for sqlite, not mysql
{% endset %}
{% set results = run_query(sql) %}
{% if results|length == 0 %}
    create schema if not exists {{ relation.without_identifier() }}
```

When the attached database type is `mysql`:

1. The query returns **no results** (it only matches `sqlite`)
2. dbt proceeds to run `CREATE SCHEMA IF NOT EXISTS <db>.<schema>`
3. DuckDB forwards this to MySQL, where `CREATE SCHEMA` ≡ `CREATE DATABASE`
4. The `IF NOT EXISTS` clause is lost during forwarding
5. MySQL errors because the database already exists

### `profiles.yml`

```yaml
outputs:
  dev:
    type: duckdb
    database: ducklake
    attach:
      - path: "mysql://user:password@localhost:3306/mydb"
        type: mysql
        alias: my_mysql_db
```

### `dbt_project.yml`

```yaml
models:
  my_project:
    reverse_etl:
      +database: my_mysql_db
      +schema: mydb
```

### Steps

1. Start a MySQL 8.0 instance with database `mydb` already created
2. Run `dbt run --select reverse_etl_model`
3. Observe the `CREATE SCHEMA` error

## Workaround

Override `duckdb__create_schema` in the project's `macros/` directory to also skip schema creation for `mysql` databases:

```sql
{% macro duckdb__create_schema(relation) -%}
  {%- call statement('create_schema') -%}
    {% set sql %}
        select type from duckdb_databases()
        where lower(database_name)='{{ relation.database | lower }}'
        and type in ('sqlite', 'mysql')
    {% endset %}
    {% set results = run_query(sql) %}
    {% if results|length == 0 %}
        create schema if not exists {{ relation.without_identifier() }}
    {% else %}
        {% if relation.schema != 'main' and results.columns[0].values()[0] == 'sqlite' %}
            {{ exceptions.raise_compiler_error(
                "Schema must be 'main' when writing to sqlite "
                ~ "instead got " ~ relation.schema
            )}}
        {% endif %}
    {% endif %}
  {%- endcall -%}
{% endmacro %}
```

## Suggested Fix

The type check in `duckdb__create_schema` should handle **all non-DuckDB attached database types** (`mysql`, `postgres`, `sqlite`) where schema creation should be skipped or handled differently. A possible approach:

```diff
- and type='sqlite'
+ and type in ('sqlite', 'mysql', 'postgres')
```

Or more robustly, skip schema creation for any attached database where `type != 'duckdb'`.

## Environment

| Component | Version |
|-----------|---------|
| dbt-duckdb | 1.10.0 |
| DuckDB | 1.4.3 |
| MySQL | 8.0 (via DuckDB `mysql` extension) |

This issue was investigated and authored by AI, but asked, reviewed and approved by human.

ernestprovo23:
## Summary
- Closes #696 — `dbt run` fails at schema creation when a MySQL (or any non-duckdb) database is attached, because `duckdb__create_schema` only special-cases `type='sqlite'`.
- Generalize the type check to skip `CREATE SCHEMA` for any attached engine where `type != 'duckdb'` (mysql, postgres, iceberg, etc).
- Keep the sqlite-specific `schema must be 'main'` guard, but gate it on a type recheck so it only fires for sqlite.
- Same shape as #299 (the original sqlite path), just broader. Precedent for conditional engine-aware logic in this macro file also at #557 and #626.

## Test plan
- [x] Extended `tests/functional/plugins/test_sqlite.py` with a regression test confirming the strict-main guard still raises for sqlite.
- [x] Added `tests/functional/plugins/test_mysql.py` mirroring `test_postgres.py`'s `@pytest.mark.skip` shape (CI-friendly, exercisable locally).
- [x] `pytest tests/functional/plugins/test_sqlite.py` passes (2 passed, 1 skipped including new test_mysql).

### DIFF
--- a/dbt/include/duckdb/macros/adapters.sql
+++ b/dbt/include/duckdb/macros/adapters.sql
@@ -3,13 +3,14 @@
     {% set sql %}
         select type from duckdb_databases()
         where lower(database_name)='{{ relation.database | lower }}'
-        and type='sqlite'
+        and type != 'duckdb'
     {% endset %}
     {% set results = run_query(sql) %}
     {% if results|length == 0 %}
         create schema if not exists {{ relation.without_identifier() }}
     {% else %}
-        {% if relation.schema!='main' %}
+        {% set attached_type = results.columns[0].values()[0] %}
+        {% if attached_type == 'sqlite' and relation.schema != 'main' %}
             {{ exceptions.raise_compiler_error(
                 "Schema must be 'main' when writing to sqlite "
                 ~ "instead got " ~ relation.schema
--- a/dbt/include/duckdb/macros/adapters.sql
+++ b/dbt/include/duckdb/macros/adapters.sql
@@ -15,6 +15,8 @@
                 "Schema must be 'main' when writing to sqlite "
                 ~ "instead got " ~ relation.schema
             )}}
+        {% elif attached_type not in ['sqlite', 'mysql'] %}
+            create schema if not exists {{ relation.without_identifier() }}
         {% endif %}
     {% endif %}
   {%- endcall -%}
--- a/tests/functional/plugins/test_mysql.py
+++ b/tests/functional/plugins/test_mysql.py
@@ -1,0 +1,56 @@
+import pytest
+
+from dbt.tests.util import run_dbt
+
+
+@pytest.mark.skip(
+    reason="""
+Skipping by default because it requires a running MySQL server reachable
+from the test process plus the duckdb 'mysql' extension installed/loaded.
+
+Exercise locally with:
+    pytest --profile=file tests/functional/plugins/test_mysql.py
+
+This test guards #696: duckdb__create_schema must skip CREATE SCHEMA for
+attached MySQL databases (type='mysql') rather than raising. Mirror of
+test_postgres.py's @pytest.mark.skip shape so CI stays green while the
+case remains exercisable locally."""
+)
+class TestMySQLPlugin:
+
+    @pytest.fixture(scope="class")
+    def profiles_config_update(self, dbt_profile_target):
+        return {
+            "test": {
+                "outputs": {
+                    "dev": {
+                        "type": "duckdb",
+                        "path": dbt_profile_target.get("path", ":memory:"),
+                        "attach": [
+                            {
+                                "path": "mysql://root:root@localhost:3306/mydb",
+                                "type": "mysql",
+                                "alias": "my_mysql",
+                            }
+                        ],
+                    }
+                },
+                "target": "dev",
+            }
+        }
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "my_model.sql": """
+                {{ config(materialized='table', database='my_mysql') }}
+                select 1 as id
+            """,
+        }
+
+    def test_mysql_attached_skips_create_schema(self, project):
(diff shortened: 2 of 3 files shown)


tests/functional/plugins/test_mysql.py
```diff
@@ -0,0 +1,56 @@
...
+                {{ config(materialized='table', database='my_mysql') }}
+                select 1 as id
+            """,
+        }
+
+    def test_mysql_attached_skips_create_schema(self, project):
+        # Before #696 this would fail because duckdb__create_schema attempted
+        # CREATE SCHEMA against the attached MySQL database. After the fix
+        # the create_schema call is skipped for any non-duckdb attached type.
+        results = run_dbt(["run"])
```
hrl20, line 65: would be good to specify a schema explicitly in this test. 
Also if the schema isn't already in the mysql db - the first dbt run could pass? maybe call run_dbt twice.

ernestprovo23: good catch on both points, addressed in a31206a.

added `schema='mydb'` to the model config so it targets the MySQL database name from the connection URL explicitly rather than falling through to the attached database default.

also restructured the test to call `run_dbt(["run"])` twice, back to back. first call exercises the initial schema-resolution path, second call exercises the case where the schema is already present in the dbt graph state, which is exactly the idempotency path the fix is meant to cover. assertions after each run.

let me know if you'd prefer a different schema name or a different assertion style after the second run.

dbt/include/duckdb/macros/adapters.sql
```diff
@@ -3,13 +3,14 @@
     {% set sql %}
         select type from duckdb_databases()
         where lower(database_name)='{{ relation.database | lower }}'
-        and type='sqlite'
+        and type != 'duckdb'
```
hrl20, line 6: left a comment here - https://github.com/duckdb/dbt-duckdb/issues/696#issuecomment-4568601051 
could this change be limited to mysql?

ernestprovo23:
makes sense, narrowed it to mysql only in 3f0206b. postgres and the other attached engines keep `create schema if not exists` since they handle it fine, only mysql skips it now while the duckdb-mysql `if not exists` bug gets sorted out upstream. sqlite keeps its main-only guard. the skip-by-default mysql test still exercises the double-run idempotency path.

hrl20 (approved):

Issue #754, opened 2026-06-01
generate_schema_name prefixes target.schema onto custom schemas for attached Iceberg REST catalog databases


malon64:
## Summary

When `dbt-duckdb` targets an attached DuckDB Iceberg REST catalog (e.g. Polaris), the default `generate_schema_name` macro produces `main_sales_gold` instead of the configured namespace `sales_gold`. The result is that dbt tries to materialise tables into a namespace that does not exist in the catalog, causing runtime failures.

---

## Environment

| Component | Version |
|---|---|
| dbt-duckdb | 1.9.6 |
| DuckDB | 1.4.x |
| DuckDB extensions | `iceberg`, `httpfs`, `parquet` |
| Iceberg catalog | Polaris REST catalog, attached as alias `polaris` |

---

## Configuration

**`profiles.yml` (relevant section):**

```yaml
outputs:
  local_runtime:
    type: duckdb
    path: /tmp/openlakeforge-sales-dbt.duckdb
    extensions: [httpfs, parquet, iceberg]
    secrets:
      - type: iceberg
        name: openlakeforge_polaris_secret
        client_id: "{{ env_var('POLARIS_DBT_CLIENT_ID') }}"
        client_secret: "{{ env_var('POLARIS_DBT_CLIENT_SECRET') }}"
        oauth2_server_uri: "{{ env_var('POLARIS_TOKEN_URI') }}"
        oauth2_scope: "PRINCIPAL_ROLE:ALL"
    attach:
      - path: lakehouse
        alias: polaris
        options:
          type: iceberg
          secret: openlakeforge_polaris_secret
          endpoint: "{{ env_var('POLARIS_REST_URI') }}"
          access_delegation_mode: none
```

**`dbt_project.yml`:**

```yaml
models:
  sales_poc:
    +database: polaris
    +schema: sales_gold
    +materialized: table
```

---

## Expected vs. Actual

| | Resolved relation |
|---|---|
| **Expected** | `"polaris"."sales_gold"."mart_sales_by_day"` |
| **Actual** | `"polaris"."main_sales_gold"."mart_sales_by_day"` |

The `main_` prefix comes from `target.schema` (which is `main` by default for DuckDB profiles). dbt's default `generate_schema_name` macro concatenates it with the custom schema: `{target.schema}_{custom_schema_name}`.

---

## Root cause

[...]

PR #755, opened 2026-06-01
fix: skip target.schema prefix for attached Iceberg REST catalog databases

malon64:
## Problem

When targeting an attached Iceberg REST catalog, the default \`generate_schema_name\` macro prepends \`target.schema\` to any custom schema name, producing \`{target.schema}_{schema}\` instead of \`{schema}\`. For an attached Iceberg catalog the schema maps to an **Iceberg namespace** — a discrete catalog object — not a local DuckDB schema. The prefixed namespace does not exist in the catalog, causing runtime failures.

See #754 for a full reproduction with profile config, \`dbt_project.yml\`, and expected vs. actual resolved relations.

## Fix

Follow the existing DuckLake pattern (\`_ducklake_dbs\` / \`is_ducklake\`):

1. **\`credentials.py\`** — populate \`_iceberg_catalog_dbs\` from \`attach\` entries whose resolved type is \`\"iceberg\"\` (via \`.type\` field or \`options\` dict).
2. **\`impl.py\`** — expose \`is_iceberg_catalog_db(database_name: str) -> bool\` decorated \`@available\`.
3. **\`macros/adapters.sql\`** — add \`duckdb__generate_schema_name\` that returns \`custom_schema_name\` unchanged when the node's configured database is an Iceberg catalog attachment, and falls back to the standard \`{target.schema}_{custom_schema_name}\` behaviour otherwise.

Project-level \`generate_schema_name\` overrides dispatch before adapter macros, so existing projects with the current workaround are unaffected.

## Tests

- 6 unit tests in \`test_credentials.py\` — \`_iceberg_catalog_dbs\` population via options dict, via \`.type\` field, path fallback when no alias, non-iceberg attach, SQLite attach, case-insensitive type matching.
- 4 unit tests in \`test_duckdb_adapter.py\` — \`is_iceberg_catalog_db\` returns true for an iceberg attach, false for a regular db name, false when no iceberg attach is configured, false for \`None\`.

All 10 new tests pass. 101 pre-existing unit tests pass.

### DIFF
--- a/dbt/adapters/duckdb/credentials.py
+++ b/dbt/adapters/duckdb/credentials.py
@@ -245,6 +245,9 @@ class DuckDBCredentials(Credentials):
         if self.is_ducklake or "ducklake:" in self.path.lower():
             self._ducklake_dbs.add(self.path_derived_database_name(self.path))
 
+        # Build set of Iceberg REST catalog database names for schema name generation
+        self._iceberg_catalog_dbs: set = set()
+
         if self.attach:
             for attachment in self.attach:
                 is_ducklake_flag = getattr(attachment, "is_ducklake", None)
@@ -259,6 +262,17 @@ class DuckDBCredentials(Credentials):
                         self._ducklake_dbs.add(alias)
                     else:
                         self._ducklake_dbs.add(self.path_derived_database_name(path))
+
+                # Detect Iceberg REST catalog via .type field or options dict
+                attachment_type = None
+                if attachment.type:
+                    attachment_type = attachment.type.lower()
+                elif attachment.options and "type" in attachment.options:
+                    attachment_type = str(attachment.options["type"]).lower()
+
+                if attachment_type == "iceberg":
+                    name = alias or self.path_derived_database_name(path)
+                    self._iceberg_catalog_dbs.add(name)
 
         # Add MotherDuck plugin if the path is a MotherDuck database
         # and plugin was not specified in profile.yml
--- a/dbt/adapters/duckdb/impl.py
+++ b/dbt/adapters/duckdb/impl.py
@@ -160,6 +160,13 @@ class DuckDBAdapter(SQLAdapter):
         return relation.database in self.config.credentials._ducklake_dbs
 
     @available
+    def is_iceberg_catalog_db(self, database_name: Optional[str]) -> bool:
+        """Return True if database_name is an attached Iceberg REST catalog."""
+        if not database_name:
+            return False
+        return database_name in self.config.credentials._iceberg_catalog_dbs
+
+    @available
     def convert_datetimes_to_strs(self, table: "agate.Table") -> "agate.Table":
         import agate
 
--- a/dbt/include/duckdb/macros/adapters.sql
+++ b/dbt/include/duckdb/macros/adapters.sql
@@ -46,6 +46,20 @@
   {{ return(run_query(sql)) }}
 {% endmacro %}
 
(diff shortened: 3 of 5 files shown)


jwills:
Hey @malon64, I don't think this is a good idea for a few reasons: first, I'm not wild about the catalog naming behavior varying based on attachment type-- like I mentioned in the discord, overwriting the `generate_schema_name` macro is basically a right-of-passage for dbt projects, and I think it's the right fix for this situation. Changing the default here would be an unpleasant surprise for anyone who has fixed this the other way-- by creating multiple catalogs in their iceberg DB based on the old classic pattern and now suddenly has data overwritten that they didn't explicitly tell us to overwrite, which is basically the worse thing that a project like this can do-- much worse IMO than just failing to do a write in the first place.

And finally, (and this is not your fault at all!) the @dataders and @hrl20 bit of this-- the way we're doing all of these various attachment-type overrides in the project right now for MD/ducklake/iceberg gives me the ick, and I have extra anxiety about introducing more of it since it seems like in #758 the fusion adapter isn't supporting the @-available annotation at the moment (or possibly going forward?) which seems like the kind of thing that would require some major surgery given how often we use this pattern in here right now.

malon64:
@jwills I agree we should not silently change the default schema behavior for all Iceberg attachments. But for REST Iceberg catalogs, schema maps to an Iceberg namespace, and dbt’s default target_schema_custom_schema behavior prevents users from targeting existing namespaces such as gold, silver, or bronze. Requiring every user to override generate_schema_name creates a poor adapter experience and is easy to get wrong.

Would you be open to an explicit opt-in profile config, disabled by default, that preserves the custom schema only for Iceberg catalog relations? That would avoid backward compatibility issues while giving Iceberg users a first-class way to target catalog namespaces.

Something like this in profile.yml
```
outputs:
  dev:
    type: duckdb
    path: dev.duckdb
    schema: main
    iceberg_schema_naming: preserve

```

jwills:
yah, that is an option (tho again, I'm not convinced it's actually materially better than simply overriding the get_schema_name, which again, is a very common thing in dbt projects.) Is this schema thing just intractably broken in Iceberg REST catalogs for some reason? I don't understand why it works this way vs. the way every other DB catalog works

malon64:
I think the issue is not that Iceberg REST catalogs are fundamentally broken, but that the usual dbt schema-generation convention becomes awkward for catalog-backed Iceberg namespaces.

In DuckDB, a fully qualified relation is catalog.schema.table. For a local DuckDB database, `main` is both commonly the default catalog and the default schema, so `target.schema = main` is natural.

But with an attached Iceberg REST catalog, the catalog is already represented by the database/attachment alias, e.g. `polaris`, and the schema part maps to the Iceberg namespace. In my case I needed to write to the existing namespace `gold`, i.e. `polaris.gold.table`.

The problem is that dbt’s default `generate_schema_name` turns `target.schema = main` + `schema = gold` into `main_gold`. So dbt-duckdb tries to write to `polaris.main_gold.table`, which is a different Iceberg namespace and not what the user configured.

I agree this should not silently change for every existing project. But from a dbt-duckdb Iceberg UX perspective, there is currently no profile-level way to say “for this attached Iceberg catalog, use the configured custom schema as the actual namespace.” The only workaround is a project macro override, which works, but feels like adapter-specific boilerplate every Iceberg user has to discover.

As I said a compromise would be to have a configuration option in the profile for iceberg users

Issue #559, opened 2025-05-30
dbt-duckdb update to work with dbt fusion engine?


ran-codes:
It would be nice to elverage the new DBT engine in duckd-dbt. Possibly related issue on their end: https://github.com/dbt-labs/dbt-fusion/issues/46

ran-codes:
opened an issue on their end just to get it on their radar. Super hyped for fusion engine and the potential to integrated with duckdb. Thank you for the great work!

eliasdefaria:
Yep confirming this is live and functional! Both SQL comprehension and the base adapter. Y'all should be good to close this now

Issue #458, opened 2024-10-10
Tolerate `grants` in models


benjaminbauer:
# Use Case
I want to use a duckdb target alongside a postgres target. One challenge: I define [grants](https://docs.getdbt.com/reference/resource-configs/grants) in my `dbt_project.yml`

# Current behaviour
I get a parser error with the duckdb target

# Desired behaviour
The duckdb target works with defined grants and ignores the grants with a warning like `grants defined on x ignored, unsupported by duckdb`

jwills:
Hey @benjaminbauer, thanks for the heads up here-- I _think_ that the change in PR #459 is all that is required here, but I'm wondering if there is any other stuff like this that would be needed for a smooth switch between Postgres and DuckDB for a project? Maybe indexes too (which aren't yet supported in dbt-duckdb for reasons, though I do periodically try to make them work every now and again)

benjaminbauer:
@jwills thanks for the prompt PR!

I am not that far into this journey. The biggest pain by far I have, are missmatches in built in functions, e.g. `to_date` in pg vs `strptime` in duckdb -> big pain but not a dbt-duckdb topic, can be solved with jinja2 and some copium.

I did not stumble over indexes so far, because dbt-duckdb does not fail when indexes are defined in dbt.

Other than that, my models build w/o problems.

A point that _could_ be a pain: [contract](https://docs.getdbt.com/reference/resource-configs/contract) enforcement and the related [constraints](https://docs.getdbt.com/reference/resource-properties/constraints). So far I only do `data_type`, `non_null`, and `unique`. All of which are supported or at least dbt-duckdb did not warn about the lack thereof

benjaminbauer:
the PR works for me btw 🚀 

I get the message `Grants are not (currently!) supported for DuckDB` as expected. Can you make that a warning?

jwills:
yes, easily! I also changed it to include the name of the relation in question, will upgrade from info -> warn in the PR

jwills:
Oops, nope, I lied-- it can only be an `info` apparently: https://docs.getdbt.com/reference/dbt-jinja-functions/log

benjaminbauer:
on second thoughts: that will probably be really noisy, because grants are often used for a whole group of models. Would it be possible to have a global warning if grants are present and not nag on each instance?

benjaminbauer:
[...]

PR #758, opened 2026-06-04
fix: replace adapter.warn_once in duckdb__apply_grants for Fusion compatibility

pgoslatara:
Under dbt-core 2.0.0 / Fusion, `dbt build` fails for any model that has a `grants` config with:

```
[JinjaError (dbt1501)]: Error executing materialization macro
'dbt_duckdb.materialization_table_duckdb' for model ...:
Failed to eval the compiled Jinja expression unknown method:
map has no method named warn_once
  (in dbt_internal_packages/dbt-duckdb/macros/adapters.sql:259:7)
  (in dbt_internal_packages/dbt-adapters/macros/adapters/apply_grants.sql:152:31)
```

The call site is `duckdb__apply_grants` in `dbt/include/duckdb/macros/adapters.sql`, which invokes `adapter.warn_once(...)`. `warn_once` is defined in `dbt/adapters/duckdb/impl.py` and is decorated with `@available`, so it works fine under the legacy Python Jinja runtime. Fusion's Minijinja-based runtime, however, does not expose `@available`-decorated adapter methods on the `adapter` proxy — it sees the proxy as a generic map and the call fails. The same model build succeeds under dbt-core 1.x.

This PR keeps the original intent of #458 (tolerate `grants` on DuckDB without crashing the parser) while making the macro work under both runtimes. `exceptions.warn` is part of the standard dbt Jinja context, so it is bound in Python Jinja and Minijinja alike.

```diff
 {% macro duckdb__apply_grants(relation, grant_config, should_revoke=True) %}
     {#-- If grant_config is {} or None, this is a no-op --#}
     {% if grant_config %}
-      {{ adapter.warn_once('Grants for relations are not supported by DuckDB') }}
+      {% do exceptions.warn('Grants for relations are not supported by DuckDB') %}
     {% endif %}
 {% endmacro %}
```

Trade-off: `exceptions.warn` does not dedup, so a project with N models configured with `grants` will see N warnings per run instead of one. That seems acceptable as the cost of cross-runtime compatibility, but happy to take a different approach (for example deleting the override entirely and falling back to the default `apply_grants`, matching `dbt-clickhouse`/`dbt-exasol`/`dbt-salesforce`) if maintainers prefer.

Refs: #458, #559.

### DIFF
--- a/dbt/include/duckdb/macros/adapters.sql
+++ b/dbt/include/duckdb/macros/adapters.sql
@@ -381,7 +381,7 @@ def materialize(df, con):
 {% macro duckdb__apply_grants(relation, grant_config, should_revoke=True) %}
     {#-- If grant_config is {} or None, this is a no-op --#}
     {% if grant_config %}
-      {{ adapter.warn_once('Grants for relations are not supported by DuckDB') }}
+      {% do exceptions.warn('Grants for relations are not supported by DuckDB') %}
     {% endif %}
 {% endmacro %}
 


jwills:
eh I'm more concerned that the fusion adapter doesn't support "at-available" since we use it all over the place in here /cc @dataders

dataders:
@pgoslatara @jwills I just opened https://github.com/dbt-labs/dbt-core/issues/15211 to discuss next steps with this! likely this solution is what we can adopt, but wanna get the ducks in a row first.

I also opened an internal thread about it

PR #760, opened 2026-06-06
docs: document the retries profile settings

### LINKED ISSUE #687: global retry setting
from time to time we have random query failures due to transient issues (network, disk, memory, ...) as one failed query completely breaks a dbt run and there is no retry logic builtin in dbt itself it would be great to have a retry setting in dbt-duckdb which would basically allow to retry the query for a defined amount with a configurable backoff.

i guess this should be possible on dbt-duckdb itself without special integration in dbt required.

Onipounds:
Documents the existing `retries` profile settings (`connect_attempts`, `query_attempts`, `retryable_exceptions`) in the README. These were previously undocumented, as noted in #687.

Closes #687

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,6 +317,32 @@ of paths on the filesystem that contain additional Python modules that should be
 can be accessed by the running dbt process and used to define custom dbt-duckdb Plugins or library code that is
 helpful for creating dbt Python models.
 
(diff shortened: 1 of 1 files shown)


jwills (approved):
LG-- thank you!

PR #763, opened 2026-06-08
Use DESCRIBE for get_columns_in_relation to avoid full catalog load (closes #762)

### LINKED ISSUE #762: duckdb__get_columns_in_relation routes temp-table introspection through global information_schema.columns, forcing a full DuckLake catalog load (726s observed)
## What happens

For an incremental model with incremental_strategy='merge', dbt-duckdb calls get_columns_in_relation on the merge staging temp table. The temp table is a DuckDB-local CREATE TEMPORARY TABLE (it has no schema/database), so the macro emits a query against system.information_schema.columns filtered only by table_name.

information_schema.columns is a global view spanning all attached catalogs. With a DuckLake catalog attached (metadata in Postgres), satisfying this view forces DuckDB to materialise DuckLake's entire column catalog from the metadata database even though the target temp table lives only in DuckDB's local temp catalog.

The same query on the target relation (which has schema + catalog in the WHERE clause) completes in 0.344 seconds.

## To reproduce

1. Attach a DuckLake catalog (Postgres metadata) with a non-trivial number of tables/columns.
2. Define an incremental model with incremental_strategy='merge'.
3. dbt run --debug and inspect the timing of the get_columns_in_relation call against the __dbt_tmp relation.

## Environment

dbt-duckdb: 1.10.1
dbt-core: 1.11.8
DuckDB: 1.5.3
Catalog: DuckLake (metadata in Postgres, data on S3, the Postgres catalog has large number of snapshots)
Incremental strategy: merge
on_schema_change: ignore

### Temp table is DuckDB-local:

```
create temporary table "imports_bronze__dbt_tmp20260608075238167429" as ( ... )
-- OK in 12.530s
```

### First call: temp table, no schema/catalog filter (726 seconds):

```
-- 06:52:50 ... SQL status: OK in 726.659 seconds
select
    column_name,
    data_type,
    character_maximum_length,
    numeric_precision,
    numeric_scale
from system.information_schema.columns
where table_name = 'imports_bronze__dbt_tmp20260608075238167429'
order by ordinal_position

```

### Second call: real table, with schema + catalog filter (0.344 seconds):

```
-- 07:04:57 ... SQL status: OK in 0.344 seconds
select
    column_name,
    data_type,
    character_maximum_length,
    numeric_precision,
    numeric_scale
from system.information_schema.columns
where table_name = 'imports_bronze'
  and lower(table_schema) = 'dev_bronze'
  and lower(table_catalog) = 'my_ducklake'
order by ordinal_position
```

I believe the difference is that temp relations have no schema or database set, so the WHERE clause contains only table_name, forcing a full scan of the Postgres-backed catalog. For contrast, the information_schema.tables lookups in the same run were 0.4–2.2s each. Only the columns view is pathological, because that's what forces DuckLake to load per-column metadata.

## Root Cause

duckdb__get_columns_in_relation ([dbt/include/duckdb/macros/adapters.sql](https://github.com/duckdb/dbt-duckdb/blob/main/dbt/include/duckdb/macros/adapters.sql)) queries the global system.information_schema.columns. For a temp relation, relation.schema and relation.database are empty, so no catalog predicate is emitted and the scan cannot be pruned to the local temp catalog.

Enumerating that view makes DuckLake materialize its whole catalog via one query (FROM ducklake_table LEFT JOIN ducklake_column for the active snapshot, with per-column correlated subqueries) against the Postgres metadata DB. See [ducklake_metadata_manager.cpp](https://github.com/duckdb/ducklake/blob/main/src/storage/ducklake_metadata_manager.cpp). On a catalog with accumulated snapshots/columns this dominates runtime. (Related: [ducklake#399](https://github.com/duckdb/ducklake/issues/399), which recommends DESCRIBE over information_schema.columns for DuckLake.)

## Suggested Fix

Resolve the relation directly instead of scanning the global view. DESCRIBE/PRAGMA table_info look the relation up via the catalog search path (local temp first) and never trigger a DuckLake catalog load. DESCRIBE returns column_name, column_type, null, key, default, extra, it just needs alias to the contract sql_convert_columns_in_relation expects:

```
{% macro duckdb__get_columns_in_relation(relation) -%}
  {%- if not relation.database and not relation.schema -%}
    {# local temp relation: avoid the global information_schema scan #}
    {% call statement('get_columns_in_relation', fetch_result=True) %}
      select
          column_name,
          column_type as data_type,
          cast(null as bigint) as character_maximum_length,
          cast(null as bigint) as numeric_precision,
          cast(null as bigint) as numeric_scale
      from (describe {{ relation.render() }})
    {% endcall %}
    {{ return(sql_convert_columns_in_relation(load_result('get_columns_in_relation').table)) }}
  {%- endif -%}
  {# existing information_schema.columns path for catalog-qualified relations #}
  ...
{% endmacro %}
```

(DuckDB's column_type already carries precision/scale in the type string, e.g. DECIMAL(18,2), so the null literals are safe for Column construction.) Could also apply DESCRIBE for all relations but its slowness seems to only be an issue for temp/unqualified case.

jwills:
## Problem

`duckdb__get_columns_in_relation` queries the global `system.information_schema.columns` view. That view spans **all** attached catalogs, so enumerating it forces a full catalog load for catalogs like DuckLake whose column metadata lives in an external metadata database. [#762](https://github.com/duckdb/dbt-duckdb/issues/762) reports this at **726s** for a single staging-table lookup during an incremental `merge`.

The reporter ([@chaegordon](https://github.com/chaegordon)) did the legwork here, including the non-obvious part: adding a `table_schema`/`table_catalog` predicate does **not** help. The first `information_schema.columns` hit on a cold connection materializes DuckLake's entire column catalog regardless of the filter — normal runs just pay that cost on the first (temp-table) lookup and the later filtered lookups only *look* fast because the catalog is already warm.

## Fix

Resolve columns via `DESCRIBE` instead. It looks the single relation up through the catalog search path and never triggers the global enumeration. Reporter confirmed the model drops from ~726s to **<15s** end to end.

DuckDB's `column_type` already carries any precision/scale/length in the type string (e.g. `DECIMAL(18,2)`), and I verified `DESCRIBE`'s `column_type` is byte-identical to `information_schema`'s `data_type` for `DECIMAL`, `VARCHAR`, `STRUCT`, nested structs, `MAP`, and lists — so the dropped `numeric_precision`/`numeric_scale`/`character_maximum_length` columns are safe to null out (DuckDB's `Column` never consulted them for these types anyway).

## Note for reviewers — the try/except

`DESCRIBE` raises a `CatalogException` for a missing relation, whereas the old `information_schema` query returned an empty list. The adapter override normalizes that back to `[]`.

This control-flow-via-exception looks a little off at first glance, but it's the **established pattern for `DESCRIBE`-based dbt adapters** — `dbt-spark`/`dbt-databricks` do the identical thing in their `get_columns_in_relation` (`impl.py`), with the comment *"spark would throw error when table doesn't exist, where other CDW would just return an empty list, normalizing the behavior here."* The alternatives are worse: a pre-check query reintroduces per-call metadata latency (the exact thing we're removing), and the in-memory relation cache isn't reliably populated for temp relations. The `[]` behavior is also relied on by dbt's shared `check_table_does_not_exist` test helper.

## Tests

Added `tests/functional/adapter/test_get_columns_in_relation.py` covering type-string fidelity (precision + nested struct) and the missing-relation → `[]` guard. Regression-ran the affected paths (incremental, external/struct-flatten, persist_docs, constraints, basic adapter-method + docs-generate): **63 passed, 4 skipped** (S3/DuckLake creds-gated).

### DIFF
--- a/dbt/adapters/duckdb/impl.py
+++ b/dbt/adapters/duckdb/impl.py
@@ -139,6 +139,16 @@ class DuckDBAdapter(SQLAdapter):
         # an unfiltered information_schema.columns scan triggers a full metastore load.
         return
 
+    def get_columns_in_relation(self, relation: BaseRelation) -> List[BaseColumn]:
+        # DESCRIBE (see issue #762) raises on a missing relation; preserve the old
+        # information_schema behavior of returning [] for callers that probe existence.
+        try:
+            return super().get_columns_in_relation(relation)
+        except DbtRuntimeError as e:
+            if "does not exist" in str(e).lower():
+                return []
+            raise
+
     @available
     def parse_index(self, raw_index: Any) -> Optional[DuckDBIndexConfig]:
         return DuckDBIndexConfig.parse(raw_index)
--- a/dbt/include/duckdb/macros/adapters.sql
+++ b/dbt/include/duckdb/macros/adapters.sql
@@ -257,24 +257,17 @@ def materialize(df, con):
 {% endmacro %}
 
 {% macro duckdb__get_columns_in_relation(relation) -%}
+  {# DESCRIBE avoids the global information_schema.columns scan, which forces a full
+     catalog load on attached catalogs like DuckLake (issue #762). DuckDB's column_type
+     carries precision/scale in the type string, so the null literals are safe. #}
   {% call statement('get_columns_in_relation', fetch_result=True) %}
       select
           column_name,
-          data_type,
-          character_maximum_length,
-          numeric_precision,
-          numeric_scale
-
-      from system.information_schema.columns
-      where table_name = '{{ relation.identifier }}'
-      {% if relation.schema %}
-      and lower(table_schema) = '{{ relation.schema | lower }}'
-      {% endif %}
-      {% if relation.database %}
-      and lower(table_catalog) = '{{ relation.database | lower }}'
-      {% endif %}
-      order by ordinal_position
-
+          column_type as data_type,
+          cast(null as bigint) as character_maximum_length,
+          cast(null as bigint) as numeric_precision,
+          cast(null as bigint) as numeric_scale
+      from (describe {{ relation.render() }})
   {% endcall %}
   {% set table = load_result('get_columns_in_relation').table %}
   {{ return(sql_convert_columns_in_relation(table)) }}
--- a/tests/functional/adapter/test_get_columns_in_relation.py
+++ b/tests/functional/adapter/test_get_columns_in_relation.py
@@ -1,0 +1,46 @@
+import pytest
+
+from dbt.tests.util import (
+    get_connection,
+    relation_from_name,
+    run_dbt,
+)
+
+# duckdb__get_columns_in_relation resolves columns via DESCRIBE (issue #762). Pin the
+# type-string fidelity (precision + nested types) and the missing-relation -> [] guard.
+
+types_model_sql = """
+{{ config(materialized='table') }}
+
+select
+    1::integer as id,
+    1.5::decimal(18, 2) as amount,
+    'x'::varchar as label,
+    cast(null as struct(fname varchar, age integer)) as nested
+"""
+
+
+class TestGetColumnsInRelation:
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {"types_model.sql": types_model_sql}
+
+    def test_describe_preserves_type_strings(self, project):
+        run_dbt(["run"])
+
+        relation = relation_from_name(project.adapter, "types_model")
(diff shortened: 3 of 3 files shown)


jwills:
I don't love it, but this is the legacy behavior for the dbt-spark/dbt-databricks adapter the agent is citing: https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/behaviors/columns.py#L34

hrl20 (approved):

PR #774, opened 2026-07-08
feat(gsheet): support service-account domain-wide delegation

banana-creator-dev:
Add optional config keys to the built-in gsheet plugin (method=service):
- impersonate: a Workspace user email to impersonate via domain-wide delegation, so the service account reads sheets on their behalf without sharing each file
- keyfile: path to the service-account JSON (defaults to gspread's default)
- scopes: OAuth scopes for the impersonated credentials (defaults read-only)

Backwards compatible: all new keys are optional; existing method=service/oauth behaviour is unchanged. Adds README docs and config-parsing tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,22 @@ modules may be initialized using an arbitrary set of key-value pairs that are de
 `config` dictionary. In this example, we initialize the `gsheet` plugin with the setting `method: oauth` and we
 initialize the `sqlalchemy` plugin (aliased as "sql") with a `connection_url` that is set via an environment variable.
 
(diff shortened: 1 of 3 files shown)


jwills:
@banana-creator-dev this is cool and makes sense; did you create a GH account just to create the PR for it? 😉

banana-creator-dev:
> @banana-creator-dev this is cool and makes sense; did you create a GH account just to create the PR for it? 😉

@jwills This is a work account :)

jwills (approved):

jwills:
@banana-creator-dev thank you so much!

PR #775, opened 2026-07-09
Skip commit when connection has no open transaction

banana-creator-dev:
## Problem

`DuckDBAdapter.commit_if_has_connection` currently calls the base implementation and catches `DbtInternalError`, logging it at **warning level with a full traceback**:

```
DuckDB adapter: Commit failed with DbtInternalError: Internal Error
  Tried to commit transaction on connection "model.elementary.dbt_sources", but it does not have one open!
Traceback (most recent call last):
  ...
```

The condition is normal, not exceptional:
- DuckDB autocommits statements against some attached catalogs (e.g. **DuckLake** with a Postgres catalog), so the transaction is already closed when dbt calls commit;
- Python models never issue `BEGIN`, so `transaction_open` is always false (the case the original quick-fix targeted).

On a multi-catalog project (local DuckDB + attached DuckLake) this prints a scary traceback for practically **every incremental/view node** — dozens of times per `dbt build` — while the models themselves succeed.

## Fix

Check `connection.transaction_open` up front and skip the commit with a `debug` log, instead of triggering the exception and catching it after the fact. Behavior is unchanged (commit still happens whenever a transaction is actually open); only the noise is gone.

Also drops the now-unused `traceback` and `DbtInternalError` imports.

## Repro

`profiles.yml` with an attached DuckLake catalog:

```yaml
attach:
  - path: "ducklake:postgres:host=... dbname=ducklake ..."
    alias: lakehouse
    is_ducklake: true
```

`dbt run` any incremental model materialized into the attached catalog → warning + traceback per node before this change, silent success after.

### DIFF
--- a/dbt/adapters/duckdb/impl.py
+++ b/dbt/adapters/duckdb/impl.py
@@ -1,5 +1,4 @@
 import os
-import traceback
 from collections import defaultdict
 from dataclasses import dataclass
 from datetime import datetime
@@ -16,7 +15,6 @@ from dbt_common.contracts.constraints import ConstraintType
 from dbt_common.dataclass_schema import dbtClassMixin
 from dbt_common.dataclass_schema import ValidationError
 from dbt_common.exceptions import DbtDatabaseError
-from dbt_common.exceptions import DbtInternalError
 from dbt_common.exceptions import DbtRuntimeError
 from dbt_common.utils import encoding as dbt_encoding
 from packaging.version import Version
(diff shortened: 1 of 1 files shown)


jwills:
Thanks for this — the change is correct and a real quality-of-life fix. I traced it against the base class: `commit()`'s `transaction_open is False` check is the *only* source of `DbtInternalError` on that path, so guarding it up front loses no error coverage and actually improves on the old code, which would have swallowed any other `DbtInternalError` too. The DuckLake/attached-catalog + Python-model reasoning in the docstring is accurate.

Two things before we merge:

1. **Add a unit test.** The existing functional suite does exercise both branches — `test_incremental.py` hits the commit path (regular models `BEGIN`, so `transaction_open` is `True`), and `test_python_model.py` / `test_ducklake_*.py` hit the new skip path (no `BEGIN`). But those tests only assert the models succeed, which was true before this PR too — so they wouldn't catch a regression of the actual behavior change here (warning+traceback → debug/skip). A small unit test that asserts (a) `transaction_open=False` → no commit issued and nothing logged above `debug`, and (b) `transaction_open=True` → base commit runs, would lock in the intent.

2. **The `code-quality` check is failing on `ruff-format`** (mypy passed). Running `pre-commit run --all-files` (the config pins ruff `v0.14.10`) and pushing the reformat should turn it green.

---
_Posted by Claude on Josh's behalf, with his signoff._

jwills (approved):
Confirmed the fix reproduces and resolves the issue: with a DuckLake catalog attached, master prints a DbtInternalError warning + traceback per materialized node on an otherwise fully-successful run; this branch produces identical results with the noise gone (and a tidy debug line under --debug). Strict improvement — checks transaction_open up front instead of provoking/catching the exception, commit still fires when a transaction is actually open. Thanks!

jwills:
thank you @banana-creator-dev !

PR #785, opened 2026-07-18
fix: merge duplicate _capabilities to restore SchemaMetadataByRelations

arijitroy003:
DuckDBAdapter defines `_capabilities` twice in impl.py. The second `CapabilityDict` assignment silently overwrites the first, causing `SchemaMetadataByRelations` support to be lost. Merged both capability entries into a single assignment.


jwills (approved):
nice fix, thank you!

PR #786, opened 2026-07-18
fix: add missing return in GSheet plugin load() for range without headers

arijitroy003:
When loading a GSheet range without headers, the `load()` method creates a new DataFrame via `df.rename().drop().reset_index()` but never returns it. The function falls through to return the original unmodified `df`, silently discarding the header rename and row drop. Added the missing `return` keyword.


jwills (approved):

arijitroy003:

PR #787, opened 2026-07-18
fix: add missing @classmethod decorator to is_cancelable methods

arijitroy003:
`LocalEnvironment.is_cancelable()` and `BuenaVistaEnvironment.is_cancelable()` use `cls` as the first parameter but lack the `@classmethod` decorator. The parent `Environment` ABC declares `is_cancelable` as a `@classmethod`, so the subclass overrides should match.


jwills (approved):

arijitroy003:

PR #796, opened 2026-07-20
Fix SQL injection in Attachment.to_sql()

arijitroy003:
## Summary

`Attachment.to_sql()` in `dbt/adapters/duckdb/credentials.py` interpolates user-supplied values directly into `ATTACH` SQL statements without escaping, creating a SQL injection vector through crafted dbt profile configuration.

**Affected surfaces:**
- `self.path` — single-quoted but unescaped; a value containing `'` breaks out of the string literal
- `self.alias` — interpolated bare (unquoted); arbitrary SQL can be appended
- Option string values — single-quoted but unescaped, same issue as path

**Fix:**
- Escape single quotes in `path` and option string values by replacing `'` with `''` (standard SQL escaping)
- Quote `alias` as a DuckDB identifier using double quotes, with embedded `"` escaped to `""`
- Add helper methods `_escape_sql_string()` and `_quote_identifier()` on the `Attachment` class

## Test plan

- [x] Added `TestAttachmentSQLInjection` class with 5 tests covering injection via path, alias, and option values
- [x] Updated existing test expectations for the new alias quoting behavior
- [x] All 34 unit tests pass (`test_credentials.py` + `test_data_path_quoting.py`)

### DIFF
--- a/dbt/adapters/duckdb/credentials.py
+++ b/dbt/adapters/duckdb/credentials.py
@@ -43,13 +43,23 @@ class Attachment(dbtClassMixin):
     # which is the case in fully managed ducklake in MotherDuck.
     is_ducklake: Optional[bool] = None
 
+    @staticmethod
+    def _escape_sql_string(value: str) -> str:
+        """Escape single quotes for use inside a SQL single-quoted string literal."""
+        return value.replace("'", "''")
+
+    @staticmethod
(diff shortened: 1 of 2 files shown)


arijitroy003:

jwills:
Closing in favor of the discussion on #797 — see there for the reasoning.

PR #795, opened 2026-07-20
Fix SQL injection in CREATE SECRET via unescaped single quotes

arijitroy003:
## Summary

`Secret._format_value()` builds SQL fragments for `CREATE SECRET` statements by interpolating values inside single-quoted strings, but never escapes embedded single quotes. A secret value such as `it's` or a crafted payload like `'); DROP TABLE secrets; --` produces malformed SQL or enables injection into the generated `CREATE SECRET` statement.

This PR fixes the vulnerability by:

- **Escaping string literals**: all string values, map keys/values, list items, and scope values now have single quotes doubled (`'` -> `''`) before interpolation, which is standard SQL escaping
- **Validating secret names**: `to_sql()` now validates the secret name against a strict identifier pattern (`^[A-Za-z_][A-Za-z0-9_]*$`), raising `ValueError` for names containing injection characters (semicolons, quotes, spaces, etc.)

## Changes

- `dbt/adapters/duckdb/secrets.py`: Added `_escape_sql_string()` helper and `_validate_identifier()` validator; applied escaping in `_format_value()` for string, dict, and list branches; applied escaping to scope values and validation to secret name in `to_sql()`
- `tests/unit/test_secrets.py`: New test file with 11 tests covering plain values, single-quote escaping, injection payload neutralization, map/list escaping, name validation, and the no-name case

## Test plan

- [x] All 11 new unit tests pass (`pytest tests/unit/test_secrets.py`)
- [x] Existing unit tests unaffected
- [ ] Verify no downstream breakage in projects using secrets with normal (quote-free) values

PR #797, opened 2026-07-20
fix: sanitize SQL identifiers and values to prevent injection

arijitroy003:
## Summary

- Escape single quotes in setting values passed to `SET` in `initialize_cursor()` to prevent SQL injection via dbt profile YAML `settings`.
- Double-quote setting keys as SQL identifiers in `initialize_cursor()`.
- Quote schema and table identifiers in `SourceConfig.table_name()` and `load_source()` `CREATE SCHEMA` to prevent identifier injection.
- Add unit tests covering all sanitization paths.

## Background

`initialize_cursor()` in `environments/__init__.py` constructs `SET {key} = '{value}'` using raw string interpolation from dbt profile YAML. A malicious or accidental single quote in a setting value (e.g. `memory_limit: "1GB'; DROP TABLE t;--"`) could break out of the string literal. Similarly, schema and table names from `SourceConfig` were interpolated unquoted into DDL statements.

## Changes

| File | Change |
|------|--------|
| `environments/__init__.py` | Double-quote keys, escape `'` in values |
| `environments/local.py` | Double-quote schema in `CREATE SCHEMA IF NOT EXISTS` |
| `utils.py` | Add `_quote_identifier()`, use it in `table_name()` |
| `tests/unit/test_sql_injection.py` | 7 unit tests covering injection vectors |

## Test plan

- [x] All 99 existing unit tests pass (0 regressions)
- [x] 7 new tests verify injection is blocked for setting keys, setting values, schema names, and table identifiers

### DIFF
--- a/dbt/adapters/duckdb/environments/__init__.py
+++ b/dbt/adapters/duckdb/environments/__init__.py
@@ -222,8 +222,11 @@ class Environment(abc.ABC):
         if creds.settings is not None:
             for key, value in creds.settings.items():
                 # Okay to set these as strings because DuckDB will cast them
-                # to the correct type
-                cursor.execute(f"SET {key} = '{value}'")
+                # to the correct type.
+                # Sanitize: quote key as identifier, escape single quotes in value.
+                safe_key = '"' + key.replace('"', '""') + '"'
+                safe_value = str(value).replace("'", "''")
+                cursor.execute(f"SET {safe_key} = '{safe_value}'")
 
         # update cursor if something is lost in the copy
         # of the parent connection
--- a/dbt/adapters/duckdb/environments/local.py
+++ b/dbt/adapters/duckdb/environments/local.py
@@ -103,7 +103,8 @@ class LocalEnvironment(Environment):
         cursor = handle.cursor()
 
         if source_config.schema:
-            cursor.execute(f"CREATE SCHEMA IF NOT EXISTS {source_config.schema}")
+            safe_schema = '"' + source_config.schema.replace('"', '""') + '"'
+            cursor.execute(f"CREATE SCHEMA IF NOT EXISTS {safe_schema}")
 
         save_mode = source_config.get("save_mode", "overwrite")
         if save_mode in ("ignore", "error_if_exists"):
--- a/dbt/adapters/duckdb/utils.py
+++ b/dbt/adapters/duckdb/utils.py
@@ -30,11 +30,17 @@ class SourceConfig:
     def __contains__(self, key):
         return key in self.meta
 
+    @staticmethod
+    def _quote_identifier(name: str) -> str:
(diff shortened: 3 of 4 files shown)


arijitroy003:

jwills:
Thanks for these three @arijitroy003 — apologies for the slow turnaround, I was out.

I'm grouping #795, #796, and this one together since they're the same theme, and leaving feedback in one place.

**On the framing:** I don't think this is actually a SQL injection vulnerability. Every value being escaped here comes from the user's own `profiles.yml` / `dbt_project.yml` / source configs — the same person who already authors and runs arbitrary SQL through dbt. No privilege boundary is crossed, so "injection" overstates it and would be misleading in a changelog.

**That said, these are real bugs** — robustness bugs, not security ones. A legitimate config value with a single quote (an S3 secret/password containing `'`, a path with an apostrophe) currently generates malformed SQL and fails. Escaping fixes that, and it's worth doing.

What I'd like to do:
1. **Consolidate into one PR.** The three touch disjoint files and merge cleanly, but each copies its own `_escape_sql_string`/`_quote_identifier`. Let's put one shared helper in `utils.py` and reuse it — and one PR is far easier to review than three near-identical ones.
2. **Reframe** the title/description as "escape single quotes in generated SQL for config values that contain quotes," not "fix SQL injection."
3. **Keep** the single-quote escaping in all three.
4. **Drop the identifier-quoting changes that alter case-sensitivity** — specifically the `SourceConfig.table_name()` change here. `table_name()` feeds `pd.read_sql_table()` in the sqlalchemy plugin, which expects a bare name and breaks with a fully-quoted `"db"."schema"."table"`. The alias double-quoting in #796 has the same concern.

Roll that into a single PR and I'll get it reviewed and merged. Thanks again!

arijitroy003:
Fair point — these values do come from the dbt profile YAML, which is author-controlled, not attacker-controlled. The escaping still prevents breakage from legitimate values containing special characters (e.g., a password with a single quote in a secret), but calling it SQL injection overstates the risk. Happy to reframe these PRs (#795, #796, this one) as input sanitization for robustness rather than security fixes, or close them if you'd prefer to keep the code simpler. What's your preference?

jwills:
Thanks, @arijitroy003.

I'm going to close all three (#795, #796, this one). Since these values all come from the user's own config, the only real breakage they prevent is a legit value containing a single quote (a secret with a `'`, a path with an apostrophe) — rare enough that I'd rather not carry the extra escaping/quoting helpers and the case-sensitivity changes that come with them.

No knock on the work — it's careful and well-tested. If someone actually hits the quote-in-a-value case, I'd happily take a minimal fix for just that (single-quote escaping, no identifier quoting). Appreciate you digging in, and hope you'll send more our way.

PR #805, opened 2026-07-24
Remove unmaintained .devcontainer config

jwills:
## What

Deletes `.devcontainer/` (the `devcontainer.json` + `Dockerfile`) and the now-dead `check-json` exclude in `.pre-commit-config.yaml` that only existed to skip the JSONC devcontainer file.

## Why

The devcontainer's VS Code settings had been stale for years — they configured **black / flake8 / autopep8 / pylint**, none of which this project uses (linting/formatting is **ruff + mypy** via pre-commit), and they relied on the `python.linting.*` / `python.formatting.*` settings namespace that the VS Code Python extension removed back in 2023. The `Dockerfile` also still referenced the retired `mcr.microsoft.com/vscode/devcontainers/python` image namespace.

Drifting that far without anyone noticing is a strong signal the container wasn't actually in use. And it doesn't buy much here: dev setup is trivial and needs no provisioned services —

```bash
pip install -e .
pip install -r dev-requirements.txt
pre-commit install
```

so the container mostly represents standing maintenance cost (base-image bumps, extension churn, keeping editor settings in sync with the real toolchain). Codespaces still works fine on a repo with no devcontainer via the default universal image.

I considered modernizing it instead, but given the lack of use and how simple the plain-venv path is, removing it is the lower-maintenance call.


cobac (approved):
🧹

PR #808, opened 2026-07-28
fix: escape single quotes in generated SQL for config values

arijitroy003:
Minimal fix for config values containing single quotes (e.g., a secret/password with an apostrophe) breaking generated SQL.

Per the discussion on #795/#796/#797, this consolidates the escaping into a single PR with just the inline single-quote doubling — no identifier quoting, no helper functions, no behavior changes.

Changes:
- secrets.py: escape quotes in secret values, map keys/values, list items, and scope values
- credentials.py: escape quotes in attachment paths and option values
- environments/__init__.py: escape quotes in setting values

Each change is a single `.replace("'", "''")` at the interpolation point.

### DIFF
--- a/dbt/adapters/duckdb/credentials.py
+++ b/dbt/adapters/duckdb/credentials.py
@@ -47,7 +47,7 @@ class Attachment(dbtClassMixin):
         # remove query parameters (not supported in ATTACH)
         parsed = urlparse(self.path)
         path = self.path.replace(f"?{parsed.query}", "")
-        base = f"ATTACH IF NOT EXISTS '{path}'"
+        base = f"ATTACH IF NOT EXISTS '{path.replace(chr(39), chr(39)*2)}'"
         if self.alias:
             base += f" AS {self.alias}"
 
@@ -107,7 +107,7 @@ class Attachment(dbtClassMixin):
                         ):
                             all_options.append(f"{key.upper()} {value}")
                         else:
-                            all_options.append(f"{key.upper()} '{value}'")
+                            all_options.append(f"{key.upper()} '{value.replace(chr(39), chr(39)*2)}'")
                     else:
                         all_options.append(f"{key.upper()} {value}")
 
--- a/dbt/adapters/duckdb/credentials.py
+++ b/dbt/adapters/duckdb/credentials.py
@@ -14,6 +14,7 @@ from dbt_common.exceptions import DbtRuntimeError
 from dbt.adapters.contracts.connection import Credentials
 from dbt.adapters.duckdb.secrets import DEFAULT_SECRET_PREFIX
 from dbt.adapters.duckdb.secrets import Secret
+from dbt.adapters.duckdb.utils import escape_sql_string
 
 
 @dataclass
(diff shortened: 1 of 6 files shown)


jwills:
Thanks @arijitroy003 — this is the right scope, and sorry for the mixed signal: my "no helpers" line in #797 was about closing the PR entirely, not a style ruling. A single shared helper is what I want. Four things and I'll merge.

**1. CI will fail as-is.** `pre-commit run --all-files` runs in `main.yml`, and ruff-format at `--line-length=99` reflows two of the new lines (`secrets.py:50` is 129 chars, `credentials.py:110` is 102).

**2. Please swap `chr(39)` for a helper in `utils.py`.** I get why you did it — you can't nest `'` inside a `"`-delimited f-string before 3.12 and we support 3.10. But six copies of `.replace(chr(39), chr(39)*2)` is hard to read and hard to change later. One function, used everywhere:

```python
def escape_sql_string(value: Any) -> str:
    """Escape a config value for use inside a single-quoted DuckDB string literal.

    Config values come from the user's own profiles.yml/dbt_project.yml, so this is
    not a trust boundary; it just keeps legitimate values that contain an apostrophe
    (a secret, a password, a path) from generating malformed SQL.
    """
    return str(value).replace("'", "''")
```

Then each site reads `f"ATTACH IF NOT EXISTS '{escape_sql_string(path)}'"`. No import cycle — `utils` doesn't import anything from the package.

**3. One site is missing:** `plugins/motherduck.py:74` is a copy of the `environments/__init__.py:226` line you fixed, iterating the same `creds.settings`. MotherDuck users still hit the bug otherwise.

**4. Tests.** The whole point is a path nobody exercises normally, so it'll regress silently without them. Worth covering: attach path, attach option value, secret scalar/map/array/scope, and a settings value through `initialize_cursor` with a mock cursor. One test that feeds the generated SQL to a live `duckdb.connect()` and reads the value back is worth more than the string comparisons.

**Optional, while you're in there:** in `credentials.py` the "already quoted" branch treats a lone `'` as a matched pair, so `option: "'"` emits `DELIM '`. A `len(stripped_value) >= 2` guard on that condition sends it to the escaping branch instead. I'd leave the pass-through itself alone — escaping pre-quoted values would break configs that rely on it today.

I have this applied locally and the suite is green, so if you'd rather I push the changes onto your branch than have you redo them, just say so.

arijitroy003:
All four addressed plus the optional fix:

1. Added `escape_sql_string()` in utils.py, used everywhere
2. Replaced all inline `chr(39)` with the shared helper
3. Added the missing motherduck.py site
4. Added tests covering attach path, option values, secret types, and a live DuckDB round-trip
5. Added `len(stripped) >= 2` guard on the already-quoted branch

Lines are within 99 chars. Thanks for the detailed review!

jwills:
looks good-- running `pre-commit run --all-files` should fix the last linting thing (did we include that in the CONTRIBUTING.md guide?)

jwills:
@arijitroy003 Thanks for the fix. The only failing check is `code-quality`: `reorder-python-imports` wants `escape_sql_string` listed before `SourceConfig` / `TargetConfig` in `dbt/adapters/duckdb/environments/__init__.py`. Could you push that small import-order update and rerun the check?

arijitroy003:
Fixed — reordered the import so `escape_sql_string` comes before `SourceConfig`/`TargetConfig` alphabetically.

jwills (approved):

PR #810, opened 2026-08-01
uv migration

jameswinegar:
## Migrate packaging to uv (hatchling + hatch-vcs, tox-uv, uv-based CI)

Replaces pbr/setup.cfg with pyproject.toml. No behavior change for
installed packages; built wheel diffed clean against the 1.10.1 release
(only master drift and pbr's non-SPDX `Apache-2` license string, now
`Apache-2.0`).

- setup.py / setup.cfg / dev-requirements.txt / MANIFEST.in / pytest.ini /
  mypy.ini replaced by pyproject.toml; dev deps become a `dev` dependency
  group (`uv sync --group dev`)
- Versioning stays tag-driven via hatch-vcs. The SemVer shim in
  `__version__.py` is retained: dbt_common.semver rejects PEP 440 dev
  versions (verified), so hatch-vcs uses `no-local-version`
- tox kept, backed by tox-uv; envs auto-provision interpreters. Nightly
  env sets `uv_seed` since uv venvs don't ship pip
- CI on astral-sh/setup-uv with caching; build/release checkouts use
  `fetch-depth: 0` for tag-based versioning; release builds with
  `uv build`, trusted publishing unchanged
- uv.lock committed for reproducible CI
- CONTRIBUTING and CHANGELOG updated

Verified: unit tests on 3.10-3.13, functional basic suite (file profile),
twine check, check-wheel-contents.


jameswinegar:
**ruff and mypy → `repo: local`**

- Both are in the dev group; running them via `uv run` pins versions from `uv.lock` only.
- Previously mypy used `language: system` with a `mirrors-mypy` `rev:` that didn't control which mypy ran. ruff drifted as a result: hook `v0.14.10`, lockfile `0.16.1`.
- mypy needs project deps importable: isolated env gives 66 unresolved imports vs 11 from the project venv. Both pass under `--ignore-missing-imports`.
- `[tool.ruff.lint] select` pinned — the rule set was implicit, so enforcement tracked whichever version ran. 0.16 defaults flag ~250 findings (mostly pyupgrade) that 0.14.10 passes. Selected set preserves current behavior.
- `pre-commit-hooks` and `reorder_python_imports` stay upstream — not project deps.

kingjaiteh:
Hi I saw this pr and decided to try and verify this myself. I Verified this on a clean machine, on the theory that a packaging migration is worth proving somewhere other than the author's box. I am new to the project, so this is verification only so not a code review.

Here's my environment: WSL2 Ubuntu 24.04.4 LTS, kernel 6.6.87.2, 12 cores, system Python 3.12.3, uv 0.12.1, uv cache wiped first.
I did a Clean clone, `gh pr checkout 810` at `ce5937e`.

Addittonally followed the rewritten CONTRIBUTING.md exactly:

`uv sync --group dev`   --> 91s cold from an empty cache, 0s warm
`uv run pre-commit install`   ran ok 
`uv run pre-commit run --all-files`  ran and all 9 hookss pass for 23s cold / 1s warm and also the tree left clean
`uv run pytest tests/unit`     --> 95 passed in 4.4s            
`uv build`     sdist and wheel both built  

uv used the existing system 3.12.3 rather than downloading its own interpreter. Resolved versions: dbt-core 1.12.0, dbt-adapters 1.24.5, dbt-common 1.38.0, duckdb 1.5.4, mypy 2.3.0, pathspec 1.0.4, pytest 9.1.1, ruff 0.16.1, pre-commit 4.6.1. Footprint is about 1.1 GB of uv cache and a 972 MB `.venv`, which might be worth a line in `CONTRIBUTING.md` for anyone on a metered connection.

On the mypy question from #730: that PR deferred mypy bumps until `dbt-common` loosened its `pathspec<0.13` pin. Which is the condition that is now met. dbt-common 1.38.0 requires `pathspec<1.1,>=0.9`, mypy 2.3.0 requires `pathspec>=1.0.0`, and pathspec 1.0.4 satisfies both, so removing the dependabot ignore here looks right. The mypy hook passed on this branch.

The rest of this is about rebasing. the master has moved to `161970c` and this branch no longer merges cleanly. I notice 3 conflicts, and here's the one I think is most important from those conflicts:

`pytest.ini` is a modify/delete conflict. This branch deletes it and moves the config into `[tool.pytest.ini_options]`, while master has since added `addopts = --strict-markers` and a `skip_database_type(database_type)` marker. Resolving in favour of the branch drops both without a diff to warn you, and `skip_database_type` is used in 10 test files under `tests/functional/adapter/`, with its definition in `tests/conftest.py`. With master's `--strict-markers` restored and this branch's marker list I noticed collection fails outright rather than warning, like:

```
ERROR collecting test_demo.py
'skip_database_type' not found in `markers` configuration option
!!!!! Interrupted: 1 error during collection !!!!!
```

That is this branch's exact marker list plus master's `addopts`, on a one-test throwaway file.

Also, happy to be wrong on any of this and would appreciate feedback. It is all from `git merge-tree origin/master HEAD` against `161970c`, not from actually running the rebase.

jameswinegar:
The MotherDuck team is going to finalize this once they cut the next release. This was just bootstrapping what could be done without release capabilities.

PR #812, opened 2026-08-03
fix: only strip query parameters from MotherDuck attach paths

k0tk267:
### Problem

`Attachment.to_sql()` strips the query string from every attach path. This was added in #444 to drop the MotherDuck token (`md:my_db?motherduck_token=...`) from the path, but there is no scheme check, so it applies to non-MotherDuck paths as well.

```python
from dbt.adapters.duckdb.credentials import Attachment

Attachment(path="postgresql://user@host/db?sslmode=require").to_sql()
# current:  "ATTACH IF NOT EXISTS 'postgresql://user@host/db'"
# expected: "ATTACH IF NOT EXISTS 'postgresql://user@host/db?sslmode=require'"
```

### Impact

- The README shows `path: postgresql://username@hostname/dbname` as the recommended form, but adding `?sslmode=require` to it is dropped with no warning or error.
- `?` is a legal character in a password. If a password contains `?`, everything from that point on is cut off and authentication fails.

### Changes

- Only strip the query string from `md:` / `motherduck:` scheme paths, and leave every other path unchanged.
- Replace `str.replace()` with a split on the first `?`.

### Tests

Added two tests to `tests/unit/test_credentials.py`:

- `test_attachment_strips_query_parameters_from_motherduck_paths` — pins the existing behavior for MotherDuck paths
- `test_attachment_preserves_query_parameters_for_non_motherduck_paths` — verifies that five non-MotherDuck paths pass through untouched

### DIFF
--- a/dbt/adapters/duckdb/credentials.py
+++ b/dbt/adapters/duckdb/credentials.py
@@ -15,6 +15,8 @@ from dbt.adapters.contracts.connection import Credentials
 from dbt.adapters.duckdb.secrets import DEFAULT_SECRET_PREFIX
 from dbt.adapters.duckdb.secrets import Secret
 
+MOTHERDUCK_SCHEMES = {"md", "motherduck"}
+
 
 @dataclass
 class Attachment(dbtClassMixin):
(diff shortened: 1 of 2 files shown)


jwills:
hey @k0tk267 -- thank you for this, this makes a ton of sense! Unfortunately I just merged the fix from PR #808 which has a merge conflict with this-- would you mind rebasing it and pushing and then I will approve it/merge it!

k0tk267:
@jwills 
Done — resolved the conflict with #808 and pushed. 
Thanks for building such a handy library, I'll definitely keep using it! 🙌

jwills (approved):

PR #813, opened 2026-08-03
Enable tests for DuckLake

cobac:
This PR adds two testing targets: local DuckLake and managed DuckLake on MotherDuck.

Apart from the tests, there are two expected differences in user-facing behavior that this PR introduces:

- The tests flagged that dbt snapshots were not working as intended on DuckLake, since their temporary-relations were not special cased as the temporary relations of incremental models were. The adapter would attempt to pass an unsupported `drop cascade` on teardown. This PR aligns how temporary relations are handled in both cases. d62f6d2702ccb6256f2faa97c6da92763368f23a
- Some invalid ducklake configurations will nwo throw compile-time errors instead of runtime errors. bbe8805db5fa23db848f4f749ec9b0fa700c9e21

### DIFF
--- a/dbt/include/duckdb/macros/adapters.sql
+++ b/dbt/include/duckdb/macros/adapters.sql
@@ -309,6 +309,9 @@ def materialize(df, con):
 {% endmacro %}
 
 {% macro duckdb__make_temp_relation(base_relation, suffix) %}
+    {# Creates duckdb temporary tables, i.e. ephemeral per connection and on the temp.main schema.
+    Temporary tables are always allocated in the DuckdDB client's memory, even when called from a MotherDuck connection. #}
+
     {% set tmp_identifier = base_relation.identifier ~ suffix ~ py_current_timestring() %}
     {% do return(base_relation.incorporate(
                                   path={
@@ -318,22 +321,31 @@ def materialize(df, con):
                                   })) -%}
 {% endmacro %}
 
-{% macro duckdb__make_temporary_relation(target_relation, batch_id='') %}
-    {% set temporary = not adapter.is_motherduck() %}
-    {% set temporary_relation = make_temp_relation(target_relation) %}
+{% macro duckdb__dispatch_temporary_relation(target_relation, batch_id='') %}
+    {# Prepares a relation that is meant to be temporary.
 
-    {# MotherDuck requires a qualified regular table because it does not support remote temp tables. #}
-    {% if not temporary %}
-        {% set temporary_relation = temporary_relation.incorporate(
+     This macro is called when creating the temporary relations needed for incremental models and snapshots.
+
+     When running against duckdb,
+     it dispatches to standard dbt-core machinery to create a temporary table.
+     However, MotherDuck doesn't support server-side temporary tables,
+     so it dispatches to a fully qualified unique table relation under a temporary schema. #}
+
+    {% if adapter.is_motherduck() %}
+        {% set relation = target_relation.incorporate(
             path=adapter.get_temp_relation_path(target_relation, batch_id)
         ) %}
-        {% do run_query(create_schema(temporary_relation)) %}
+        {% do run_query(create_schema(relation)) %}
         {% if not adapter.disable_transactions() %}
             {% do adapter.commit() %}
         {% endif %}
+        {% do return({'relation': relation, 'temporary': false}) %}
+    {% else %}
+        {% do return({
+            'relation': make_temp_relation(target_relation),
+            'temporary': true
+        }) %}
     {% endif %}
-
-    {% do return({'relation': temporary_relation, 'temporary': temporary}) %}
 {% endmacro %}
 
 {% macro duckdb__current_timestamp() -%}
--- a/dbt/include/duckdb/macros/adapters.sql
+++ b/dbt/include/duckdb/macros/adapters.sql
@@ -318,6 +318,24 @@ def materialize(df, con):
                                   })) -%}
 {% endmacro %}
 
+{% macro duckdb__make_temporary_relation(target_relation, batch_id='') %}
+    {% set temporary = not adapter.is_motherduck() %}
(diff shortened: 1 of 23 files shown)


cobac:
The MD DuckLake tests are on the same concurrency group as the rest of MD, let me see how fast things run on CI. We could consider using more than one duckling for CI...

tox.ini
```diff
@@ -78,3 +78,23 @@ commands =
...
   -rdev-requirements.txt
   -e.
+
+[testenv:ducklake]
+description = adapter functional testing using a local DuckLake database
+skip_install = True
+passenv = *
+commands = {envpython} -m pytest --database-type=ducklake {posargs} tests/functional/adapter
+deps =
+  duckdb==1.5.4
```
hrl20: i think we should be able to leave this unpinned (since it doens't test motherduck)

tests/ducklake.py
```diff
@@ -0,0 +1,29 @@
...
+    if profile_type == "memory":
+        root = Path(tmpdir_factory.mktemp("ducklake"))
+        profile["path"] = f"ducklake:{root / 'catalog.ducklake'}"
+    else:
+        profile["is_ducklake"] = True
+
+
+def create_motherduck_database_sql(database_name: str, database_type: str) -> str:
+    if database_type == "ducklake":
+        return f"CREATE DATABASE {database_name} (TYPE ducklake)"
```
hrl20: any reason not do `IF NOT EXISTS` here?

tests/conftest.py
```diff
@@ -142,6 +160,20 @@ def skip_by_profile_type(profile_type, request):
                 pytest.skip(f"skipped on '{profile_type}' profile")
 
 
+@pytest.fixture(autouse=True, scope="class")
+def skip_by_database_type(database_type, request):
+    for marker in request.node.iter_markers("skip_database_type"):
+        if database_type in marker.args:
+            pytest.skip(f"skipped on '{database_type}' database type")
```
hrl20: is this one needed?

dbt/include/duckdb/macros/materializations/incremental.sql
```diff
@@ -36,16 +32,11 @@
   {{ drop_relation_if_exists(preexisting_backup_relation) }}
 
   {% set to_drop = [] %}
+  {%- set batch_id = adapter.batch_id_for_model(model) -%}
+  {% set temporary_relation = duckdb__make_temporary_relation(target_relation, batch_id) %}
+  {% set temp_relation = temporary_relation.relation %}
+  {% set temporary = temporary_relation.temporary %}
   {% if not temporary %}
-    -- if not using a temporary table we will update the temp relation to use a different temp schema ("dbt_temp" by default)
-    -- for microbatch with concurrent batches, include batch timestamps in the identifier to avoid collisions
```
hrl20, line 41: let's keep this comment in

dbt/include/duckdb/macros/adapters.sql
```diff
@@ -318,6 +318,24 @@ def materialize(df, con):
                                   })) -%}
 {% endmacro %}
 
+{% macro duckdb__make_temporary_relation(target_relation, batch_id='') %}
```
hrl20: note: there's also `duckdb__make_temp_relation`, so this new function name is a bit confusing. 
i now remember trying to fix the snapshot behavior once just inside duckdb__make_temp_relation-  https://github.com/duckdb/dbt-duckdb/pull/686/changes i thought it was roughly the right approach for the snapshot issue, but it didn't fully work and i didn't investigate further. do you think there's a way to make this work with just changes to `duckdb__make_temp_relation`?

hrl20 (commented):
thanks for setting this up!!

tests/ducklake.py
cobac: hmmm not really, although I see that's what we use for duckdb databases

we use random db names, and tear them down... if the pytest process doesn't die in the way

I think probably what we want is `create or replace` in the unlikely case there's a collision? I'll change both to `create or replace`, let me know if I'm missing smth

tests/conftest.py
cobac: Not needed-needed, but it's needed to avoid initializing unused fixtures the way we're marking skipping. Without it, smth like 

```python
@pytest.mark.skip_database_type("ducklake")
class TestFeatureNotSupportedOnDucklake:
    def test_one(self):
        ...

    def test_two(self):
        ...
```

would still initialize all fixtures that apply at the class level (explicitly depended on or via `autouse`), some our fixtures are heavy-ish setting custom profiles or db state. The explicit class marker let's pytest skip before evaluating class-level fixtures.

--- 

I've refactored this bit on fd44a54ae791a24036057972296f2c5f1cdcb7c4, now I think what's going on is more explicit (we rely less on understanding magical pytest internals), and it allows to validate that there are no typos on the skip marker calls instead of failing silently.

dbt/include/duckdb/macros/adapters.sql
cobac: I've renamed the macro and clarified the dispatch.

I ·think· we can simplify this, but I'll leave that for another PR. The best way of simplifying this would be to make server-side temp tables on MD GA though 🙃

tox.ini
```diff
@@ -78,3 +78,22 @@ commands =
...
+  -rdev-requirements.txt
+  -e.
+
+[testenv:md-ducklake]
+description = adapter functional testing using a managed DuckLake database in MotherDuck
+skip_install = True
+passenv = *
+commands = {envpython} -m pytest --profile=md --database-type=ducklake --maxfail=2 {posargs} tests/functional/plugins/motherduck tests/functional/adapter
+deps =
+  duckdb==1.5.4
```
nicku33:
```suggestion
  duckdb==1.5.5
```

hrl20 (approved):

PR #815, opened 2026-08-05
add drop relation tests

cobac:
DuckDB 2.0 comes with triggers, which are dependents of tables.

Also, even currently before 2.0, setting `enable_view_dependencies=true` or `enable_macro_dependencies=true` can add dependencies to tables.


tests/functional/adapter/test_drop_relation.py
```diff
@@ -0,0 +1,133 @@
+import duckdb
+"A bit verbose, but this is one of the main areas where the accepted syntax for DuckDB databases and DuckLake differ."
```
hrl20: is this a stray comment?

hrl20 (commented):
tests looks good, though i'm not sure what you mean in the PR comment
> Also currently enable_view_dependencies=true and enable_macro_dependencies=true
break this assumption.

what's this assumption?

cobac:
reworded PR description, it was unclear "Also, even currently before 2.0, setting enable_view_dependencies=true or enable_macro_dependencies=true can add dependencies to tables."

tests/functional/adapter/test_drop_relation.py
cobac: I think I meant it as a module comment, should have been over the `import duckdb` line, which my editor added after the fact when I implemented `assert_drop_without_cascade_fails`. I'll drop it

hrl20 (approved):

PR #816, opened 2026-08-05
support DuckLake partition functions

cobac:
Currently we pass the contents of partition_by for sql escaping, assuming it's always a column identifier.
However ducklake allows for several partition functions, which no longer are interpreted as such if escaped.
https://ducklake.select/docs/stable/duckdb/advanced_features/partitioning

- day/hour/month/year(ts) etc time expressions
- bucket(n, col) hash bucketing

This change detects bypasses escaping leaving it to the user if the column inside the function requires it.

Original PR: https://github.com/motherduckdb/dbt-duckdb/pull/31

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/tests/functional/adapter/test_ducklake_partitioned_by_integration.py
+++ b/tests/functional/adapter/test_ducklake_partitioned_by_integration.py
@@ -67,10 +67,12 @@ models__empty_partitioned_by_list = """
 select 1 as ds, 'a' as value
 """
 
-models__invalid_partitioned_by_string = """
-{{ config(materialized='table', database='ducklake_db', partitioned_by='ds); drop table x; --') }}
+models__transform_partitioned_model = """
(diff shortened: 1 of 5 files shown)


tests/functional/adapter/test_ducklake_partitioned_by_integration.py
```diff
@@ -101,6 +103,33 @@ def get_partition_columns(project, model_name, schema_name):
     return [row[0].lower() for row in project.run_sql(query, fetch="all")]
 
 
+def get_partition_columns_with_transforms(project, model_name, schema_name):
```
hrl20, line 79: this looks identical to get_partition_columns except for the selected column, can we merge the two?

hrl20 (commented):
thanks!

hrl20 (approved):

PR #817, opened 2026-08-05
restore transactions for DuckLake tables

cobac:
We avoided transactions for DuckLake table materializations due to a bug where both ALTER statements on tables, where setting partition or sort by would get lost if the table was renamed within a transaction. The core issue was recently fixed in https://github.com/duckdb/ducklake/pull/1154, which went out with DuckDB 1.5.3, so we can be back to normal.

There is a workaround where you could just rerun the ALTER command post commit to recover, so for pre 1.5.3 doing that instead of entirely abandoning transactions.

There is another workaround this undoes: to avoid excessive conflicts,perist_docs was previously moved to a post tx action. However https://github.com/duckdb/ducklake/pull/1163 fixes the unnecessary conflicts so that also returns to normal post 1.5.3.

I also enabled the partition and sort key integration tests to run against the md MotherDuck profile.

Original PR: https://github.com/motherduckdb/dbt-duckdb/pull/32


tests/functional/plugins/test_sqlite.py
```diff
@@ -20,11 +20,11 @@
 """
 
 
+@pytest.mark.skip_profile("md")
```
hrl20: could you make a note on why this test was skipped?

hrl20 (approved):

tests/functional/plugins/test_sqlite.py
hrl20: i checked with @nicku33 about the reason behind this. This test is not currently included under the motherduck test environment (it only covers `tests/functional/plugins/motherduck tests/functional/adapter`)- so was never run in CI. but if running the test locally with profile=md, this test would fail (documentned why in the comment)

PR #835, opened 2026-08-24
fix(tests): make resource module import and setrlimit conditional for Windows compatibility

AhmadBilalDSA:
## Description
The top-level `import resource` in `tests/conftest.py` causes test discovery and collection to fail on Windows environments with `ModuleNotFoundError: No module named 'resource'`, as the `resource` module is Unix-specific.

This PR wraps the `resource` import and subsequent file limit adjustment (`resource.setrlimit`) in a `try...except ImportError` check. This allows the test suite to be collected and executed seamlessly across both Windows and Unix systems.

## Checklist
- [x] Made `resource` import and `setrlimit` call conditional in `tests/conftest.py`
- [x] Passed `ruff check` and `ruff format`
- [x] Verified unit tests passing locally (`113 passed`)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,10 @@
 import os
 import random
-import resource
+
+try:
+    import resource
+except ImportError:
+    resource = None
 import string
 import subprocess
 import time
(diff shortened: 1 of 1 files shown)


jwills (approved):
LG- thank you @AhmadBilalDSA , and sorry for the trouble on Windows!

PR #837, opened 2026-08-25
Fix model-level foreign key rendering

### LINKED ISSUE #836: model level foreign key constraints include database (file) name
## Setup

| package       | version |
| ------------ | ------- |
| dbt-duckdb | 1.10.1    |
| duckdb        | 1.5.5 |
| dbt-core | 1.11.13 |

## Issue Description

I'm using [dbt Contracts](https://docs.getdbt.com/docs/mesh/govern/model-contracts?version=2) with their native constraints. So far, I've used foreign key constraints only on the column-level, ie:

```yaml
version: 2
models:
  - name: my_model
    config:
      contract:
        enforced: true
    columns:
      - name: fk_col_id
        constraints:
          - type: foreign_key
            to: ref('my_other_model')
            to_columns:
              - id
        data_type: uuid
```

The adapter renders something like this:

```sql
create table "db_name"."main"."my_model__dbt_tmp" (
  fk_col_id uuid reference "main"."my_other_model" (id)
);
```

The adapter does not include the database name in the foreign key constraint.

Multi-column constraints need to be defined in the `constraints` section of the model, ie:

```yaml
version: 2
models:
  - name: my_model
    config:
      contract:
        enforced: true
    constraints:
        columns:
          - grain_part_1
          - grain_part_2
        to: ref('my_other_model')
        to_columns:
          - grain_part_1
          - grain_part_2
    columns:
      - name: grain_part_1
        data_type: uuid
      - name: grain_part_2
        data_type: uuid
```

These constraints are not added to the column itself, but added "under" the column definition:

```sql
create table "db_name"."main"."my_model__dbt_tmp" (
  grain_part_1 uuid,
  grain_part_2 uuid,

foreign key (grain_part_1, grain_part_2) references "db_name"."main"."my_other_model" (grain_part_1, grain_part_2)
);
```

So, this time the adapter added the database name in and I get an error

> Parser Error: FOREIGN KEY constraints cannot be defined cross-database

I've found some comment in the code regarding column level constraints but don't really know how this applies to model level constraints: https://github.com/duckdb/dbt-duckdb/blob/master/dbt/adapters/duckdb/impl.py#L393-L410# 

## Expected Behaviour

The same definition works for PostgreSQL so I'd expect the same behaviour for the duckdb-adapter. The definition also follow the dbt documentation: https://docs.getdbt.com/reference/resource-properties/constraints?version=2#defining-constraints.

jwills:
## Summary

- omit the database/catalog qualifier when rendering structured model-level foreign keys
- share the DuckDB-specific target normalization with column-level foreign keys
- add unit and functional coverage for composite foreign keys

Fixes #836.

## Testing

- `uv run pytest -q tests/unit` (115 passed)
- `uv run pytest -q tests/functional/adapter/test_constraints.py` (18 passed)
- `uv run pre-commit run --files dbt/adapters/duckdb/impl.py tests/unit/test_duckdb_adapter.py tests/functional/adapter/test_constraints.py`

### DIFF
--- a/tests/functional/adapter/test_constraints.py
+++ b/tests/functional/adapter/test_constraints.py
@@ -10,6 +10,7 @@ from dbt.tests.adapter.constraints.test_constraints import (
     BaseIncrementalConstraintsRollback,
     BaseModelConstraintsRuntimeEnforcement,
 )
+from dbt.tests.util import run_dbt
 
 
 pytestmark = pytest.mark.skip_database_type(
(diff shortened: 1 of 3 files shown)


tests/functional/adapter/test_constraints.py
```diff
@@ -103,3 +104,48 @@ class TestModelConstraintsRuntimeEnforcement(
     @pytest.fixture(scope="class")
     def expected_error_messages(self):
         return ["NOT NULL constraint failed"]
+
+
+@pytest.mark.skip_profile("md", "buenavista")
```
hrl20, line 109: i tried and the test works with md too.

hrl20 (approved):

PR #838, opened 2026-08-27
Support motherduck_token as a top-level profile property

Donnype:
Add a top-level `motherduck_token` credential into settings in `__post_init__` and crash on a conflict with the old settings-setup. Also drop MotherDuck config keys from the settings after moving them into the connect config: `SET motherduck_token='...'` on a live connection fails with 'can only be set during initialization' when cursors re-apply settings.


jwills:
@hrl20 sending this your way-- this was a new one to me, did this change recently?

PR #841, opened 2026-09-02
Fix sql_header on models with enforced contracts

### LINKED ISSUE #515: [bug] sql_header and contract enforcement causes adapter syntax erorrs
Using contract enforcement and setting a sql_header can cause the DuckDB adapter to encounter errors when it attempts to describe the structure of a query because the sql_header is included when it shouldn't be.

For example it attempts to run:
```yaml
# dbt_project.yml
    +sql_header: |
        set pg_connection_limit = 3 ;
        set threads = 2 ;
```

```sql
DESCRIBE (set pg_connection_limit = 3 ;
set threads = 2 ;select * from (
        

select *, current_timestamp::timestamptz as updated_at
```
Which results in the error:
```
  Parser Error: syntax error at or near "set"
```

Instead the adapter should not be inserting sql_header into the top of the query it is attempting to describe. You can work around this with:
```yaml
    +sql_header: |
      {% if execute %}
        set pg_connection_limit = 3 ;
        set threads = 2 ;
      {% endif %}

```

Donnype:
The contract schema check renders get_empty_subquery_sql and wraps the result in DESCRIBE (...). The default macro inlines the sql_header into that string, which DuckDB cannot parse (DESCRIBE takes a single query, not a statement list), so any model combining sql_header with an enforced contract fails with a parser error.

Solution: override duckdb__get_empty_subquery_sql to execute the header as its own statement via run_query instead of concatenating it. The header runs on the same connection, so session effects (SET, temp macros) the model depends on stay visible to the describe.

Fixes #515.

### DIFF
--- a/tests/functional/adapter/test_sql_header.py
+++ b/tests/functional/adapter/test_sql_header.py
@@ -1,0 +1,44 @@
+import pytest
+
+from dbt.tests.util import run_dbt
+
+model_sql = """
+{{ config(materialized='table') }}
+{% call set_sql_header(config) %}
+create or replace temp macro magic() as 42;
+{% endcall %}
+select magic() as answer
+"""
+
+contract_schema_yml = """
+models:
+  - name: header_model
+    config:
+      contract:
+        enforced: true
+    columns:
+      - name: answer
+        data_type: integer
+"""
+
+
+class BaseSqlHeader:
+    def test_sql_header(self, project):
+        run_dbt(["run"])
+        rows = project.run_sql("select answer from {schema}.header_model", fetch="one")
+        assert rows[0] == 42
+
+
+class TestSqlHeader(BaseSqlHeader):
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {"header_model.sql": model_sql}
+
+
+class TestSqlHeaderWithContract(BaseSqlHeader):
+    """sql_header used to break enforced contracts: the header was inlined into the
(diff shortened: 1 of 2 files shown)


tests/functional/adapter/test_sql_header.py
```diff
@@ -0,0 +1,44 @@
+import pytest
+
+from dbt.tests.util import run_dbt
+
+model_sql = """
+{{ config(materialized='table') }}
+{% call set_sql_header(config) %}
+create or replace temp macro magic() as 42;
```
jwills, line 8: so this doesn't work for arbitrary sql in the header tho, right? If you change this to be `create temp macro magic() as 42` the run will fail b/c the header is going to get executed twice?

Donnype: Correct as far as I can tell. I'm digging into what's expected here and I'm getting the sense that the implicit assumption for the `sql_header` is that it's idempotent (at least [the examples](https://docs.getdbt.com/reference/resource-configs/sql_header?version=2) are). I tried this out on `dbt-postgres` and saw the header run twice as well. Thoughts?

jwills: yep took a look and it seems like this is how it's done, which surprised me a bit! i'm good to approve and merge, thank you @Donnype !

jwills (approved):

PR #842, opened 2026-09-04
Add ducklake plugin to register external files in a DuckLake catalog

Donnype:
Adds a built-in plugin that registers the parquet file written by an external materialization into an attached DuckLake catalog via `ducklake_add_data_files` instead of copying the data. The lake table is created from the parquet schema if it does not exist yet. Configured with the lake's attach alias ('database' plugin config or 'ducklake_database' model config), optional 'ducklake_schema', 'ducklake_table', and pass-through of the `ignore_extra_columns` / `allow_missing_columns` options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,6 +494,39 @@ relative to the current working directory, but you can change the default direct
 
 Unfortunately incremental materialization strategies are not yet supported for `external` models.
 
+##### Registering external files in a DuckLake catalog
+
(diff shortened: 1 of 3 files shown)


jwills:
@Donnype whats the motivation for this one as plugin we ship? I think I understand the use case but as I don't really run lakehouses myself these days I may not fully grok the impact of this for other users and why it would be good to ship with it included in the core project

PR #847, opened 2026-09-17
Fix TypeError when creating the postgres plugin

dajiaohuang:
## Summary

`BasePlugin.create()` always constructs plugins with the `credentials` keyword, but the postgres plugin's `__init__` override predates that argument and does not accept it. Any profile containing a `postgres` plugin therefore fails while the DuckDB connection is being set up:

```
TypeError: Plugin.__init__() got an unexpected keyword argument 'credentials'
```

`postgres` is the only plugin module that overrides `__init__`, so it is the only affected plugin.

## Reproduction

`profiles.yml`:

```yaml
pgproj:
  target: dev
  outputs:
    dev:
      type: duckdb
      path: ":memory:"
      plugins:
        - module: postgres
          config:
            dsn: "postgresql://user@localhost:5432/mydb"
```

`dbt debug` before this change:

```
Registered adapter: duckdb=1.11.1-dev15
Connection test: [ERROR]
...
  >Plugin.__init__() got an unexpected keyword argument 'credentials'
```

After this change the plugin is constructed and the connection proceeds to the `ATTACH`:

```
DuckDB_PostgresPlugin adapter: PostgreSQL extension installed and loaded.
Connection test: [ERROR]
```

The remaining error there is the absence of a reachable Postgres server in the environment used to reproduce, not the plugin.

## Changes

- `dbt/adapters/duckdb/plugins/postgres.py`: accept `credentials` and forward it to `BasePlugin.__init__`.
- `tests/unit/test_plugins.py`: new unit tests. `test_create_forwards_credentials` builds the plugin through `BasePlugin.create()` exactly as `initialize_plugins()` does, and `test_plugin_init_accepts_create_keywords` asserts that every importable built-in plugin accepts the `name`/`plugin_config`/`credentials` keywords, so the same mismatch cannot reappear in another plugin.

Both tests fail on `master` with the `TypeError` above and pass with the fix.

## Testing

- `pytest tests/unit` — 120 passed, 5 skipped.
- `ruff check`, `ruff format`, `reorder-python-imports`, `mypy` — clean.


jwills (approved):
@dajiaohuang hey, thank you for the fix!

PR #850, opened 2026-09-21
fix: preserve attached-catalog USE across cursors so FK constraints resolve

### LINKED ISSUE #623: Foreign key reference fails with "Table does not exist" in dbt-duckdb
I’m running into an issue when trying to set up a simple pair of models with constraints defined in schema.yml. Even though random_model_2 builds successfully, random_model_1 fails with a runtime error saying the table does not exist.

Repo structure

models/random_model/ has 3 files:

schema.yml
```
version: 2

models:
  - name: random_model_2
    config:
      contract:
        enforced: true
    columns:
      - name: col
        data_type: int
        constraints:
          - type: primary_key
      - name: extra
        data_type: varchar

  - name: random_model_1
    config:
      contract:
        enforced: true
    columns:
      - name: col
        data_type: int
        constraints:
          - type: foreign_key
            expression: "elis_dev.random_model_2 (col)"
      - name: extra
        data_type: varchar
```

random_model_1.sql
```
-- depends_on: {{ ref("random_model_2") }}

select
    unnest([1, 1]) as col, unnest(['a', 'b']) as extra
```

random_model_2.sql
```
select
    unnest([1,2]) as col, unnest(['blah', 'blah']) as extra
```

When I run `dbt run --select random_model` I get the following error:

```
Failure in model random_model_1 (models/random_model/random_model_1.sql)
  Runtime Error in model random_model_1 (models/random_model/random_model_1.sql)
  Catalog Error: Table with name random_model_2 does not exist!
  Did you mean "dbname.elis_dev.random_model_2"?
```

Full logs:

```
16:23:38  1 of 2 OK created sql table model dbname.elis_dev.random_model_2
16:23:38  2 of 2 ERROR creating sql table model dbname.elis_dev.random_model_1
...
Catalog Error: Table with name random_model_2 does not exist!
Did you mean "dbname.elis_dev.random_model_2"?
```

Environment
```
python = "^3.11"
dbt-duckdb = "^1.9.2"
```
Logs show me that I'm running with `dbt=1.10.10` and `duckdb=1.9.4`

profiles.yml
```
dbt_setup:
  outputs:
    dev:
      schema: elis_dev
      threads: 1
      type: duckdb
      attach:
        - path: "postgresql://{{ env_var('DB_USER') }}:{{ env_var('DB_PASSWORD') }}@{{ env_var('DB_HOST') }}:{{ env_var('DB_PORT') }}/dbname"
          type: postgres
          alias: dbname
```

What I'm expecting here is that random_model_2 builds first, and then random_model_1 should successfully build and apply the foreign key constraint. But random_model_1 fails with a Catalog Error saying the table does not exist, even though random_model_2 exists in the same schema.

Is this a bug with how dbt-duckdb handles constraints in schema.yml, or am I misconfiguring something? Is it something to do with the Postgres attach? Is there any work around for this?

bhargaviHQ:
## Problem

When `database` in the profile is set to match one of the `attach` aliases (routing model materialization into that attached database instead of the primary one) a model with a `FOREIGN KEY` constraint fails with: 
Catalog Error: Table with name X does not exist!

## Root cause

DuckDB rejects a catalog-qualified name in a `FOREIGN KEY` constraint even one matching the *current* catalog, so a model's FK target can only resolve if the attach alias is already the connection's default catalog. A DuckDB cursor doesn't inherit a `USE` issued on its parent connection, so every new cursor defaulted back to the primary (path-derived) catalog instead of the alias, and unqualified FK targets failed to resolve.

## Fix

Re-issue `USE "<database>"` on both the base connection (`initialize_db`) and every new cursor (`initialize_cursor`) when `database` matches an attach alias.

Fixes #623

## Tests
- Added `TestForeignKeyOnAttachedDatabaseAlias` in `tests/functional/adapter/test_attach.py`, reproducing the issue with two local DuckDB files (no external database needed)
- Confirmed the test fails without the fix (`Catalog Error: Table with name fk_target does not exist!`) and passes with it
- Full functional adapter suite passes: `pytest tests/functional/adapter/ --profile file` (196 passed)
- `pre-commit run --all-files` passes


jwills:
hey @bhargaviHQ thanks for this, but it's not the right place for this to go-- will post a PR that does what I think needs doing tho

PR #851, opened 2026-09-21
fix: bind foreign keys in target catalog

### LINKED ISSUE #623: Foreign key reference fails with "Table does not exist" in dbt-duckdb

jwills:
## Problem

DuckDB rejects catalog-qualified foreign key targets. When a contracted model is routed to an attached database, its schema-qualified foreign key target is resolved against the cursor's primary catalog instead of the catalog where the model is being created.

## Fix

Select the target relation's catalog immediately before creating an enforced, non-temporary table. This keeps catalog selection scoped to the model being materialized and supports per-model database routing instead of relying on the profile-level database.

Fixes #623.

## Tests

- Added a functional regression covering two contracted models routed to an attached DuckDB database with model-level database configuration
- Verifies that the foreign key exists in the attached database
- Full file-profile adapter suite: 195 passed, 14 skipped
- pre-commit run --all-files


Voyagerroc-Lab (commented):
Review summary

The catalog switch is applied only for enforced, non-temporary tables that are materialized into an attached database, which matches the failure mode described. The functional regression uses two contracted models routed to the attached catalog and verifies the actual FOREIGN KEY constraint there. The implementation is focused; the broad Python-version and functional matrix is still running.

I found no blocking issue in the reviewed diff. I attempted an approval, but the current review token does not have explicit repository approval permission; this comment records the completed review.

